PR #3, opened 2019-05-14
Add storybook with custom theme that uses our colors

daniman:
The default result of installing Storybook with a custom theme that uses colors from our color import file.

Storybook is currently empty of any components. You can run it by running `npm run storybook`.

### DIFF
--- a/colors.js
+++ b/colors.js
@@ -28,14 +28,28 @@ const indigo = {
   lightest: '#ebe6ff'
 };
 
+const black = {
+  darker: '#12151A',
+  dark: '#14171C',
+  base: '#191C23',
+  light: '#22262E',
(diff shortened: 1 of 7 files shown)


colors.js
```diff
@@ -28,14 +28,28 @@ const indigo = {
...
+  light: '#959DAA',
+  lighter: '#B2B9C3'
+};
+
+const silver = {
+  darker: '#CAD0D8',
+  dark: '#DEE2E7',
+  base: '#EBEEF0',
+  light: '#F4F6F8',
+  lighter: '#FCFDFF'
```
trevorblades: Good call! I originally tried to smoosh these into one `grey` key to match Tailwind's default palettes, but this is the right way to go 👍

PR #5, opened 2019-05-16
Add color stories and their supporting components

daniman:
Added stories for our colors to mimic the Zeplin story for Space Kit's colors.

Added a `font.css` file to help define the font sizes in the color stories, but we don't need to export these specific fonts to all our properties. I copied them from Space Kit's typography sheet for Engine UI's font sizes.

All the styles for the stories that didn't come from `colors.js` or `font.css` are inline for now, because I wasn't sure which CSS-in-JS provider to use or how to set that up personally.


trevorblades:
@daniman looks good! Merge away 🚢

PR #8, opened 2019-06-27
AP-542: Implement Typography

justinanastos:


justinanastos:
> Looking good @justinanastos! I noticed that the typography definitions for our base and mono families are identical, except for the `fontFamily` property in each ones' `defaultTypographyDefinitions`, and the absence of `xlarge` and larger in the monospace ones.
> 
> Do you think it would be worthwhile to write these definitions in a way that allows these two files to share more code between them? We could even extend the `base` styles in the `mono` config like this:
> 
> ```ts
> import * as typography from './base';
> import {mono as fontFamily} from '../fonts';
> 
> // spread base styles in but replace the font family property with the mono one
> export const large: TypographyDefinition = {
>   ...typography.large,
>   fontFamily
> };
> 
> export const base: TypographyDefinition = {
>   ...typography.base,
>   fontFamily
> };
> ```

That's a great idea! I'll do that.

src/typography/base.ts
```diff
@@ -0,0 +1,59 @@
...
+  ...defaultTypographyDefinitions,
+  fontSize: 13,
+  lineHeight: 1.54
+};
+
+export const xsmall: TypographyDefinition = {
+  ...defaultTypographyDefinitions,
+  fontSize: 11,
+  lineHeight: 1.55,
+  textTransform: "uppercase"
```
adamatorres: @justinanastos -- I wonder if we should take out the uppercase from this since that's such a stylistic choice? It seems like more of a style guide thing than a design system thing.

src/typography/mono.ts
```diff
@@ -0,0 +1,38 @@
...
+  ...defaultTypographyDefinitions,
+  fontSize: 13,
+  lineHeight: 1.54
+};
+
+export const xsmall: TypographyDefinition = {
+  ...defaultTypographyDefinitions,
+  fontSize: 11,
+  lineHeight: 1.55,
+  textTransform: "uppercase"
```
adamatorres: @justinanastos -- same here, removing "uppercase" probably makes sense.

src/typography/base.ts
justinanastos: Totally works for me @adamatorres !

PR #10, opened 2019-06-27
Use TypeScript for colors

justinanastos:
This won't only give us type safety, but auto-complete :-D


justinanastos:
Saving this for later.

PR #11, opened 2019-06-27
Add apollo logo

mayakoneval:
adds the apollo logo to space kit


icons/src/svgs/theme/apollo_logo.svg
```diff
@@ -3,7 +3,7 @@
 <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
 	 viewBox="0 0 1024 353.9" style="enable-background:new 0 0 1024 353.9;" xml:space="preserve">
 <style type="text/css">
-	.st0{stroke:#FFFFFF;stroke-miterlimit:10;}
```
justinanastos: We'll have to change this to make it work for us. We use `svgo` to process the SVG before we create a React component with it. One of the things we do in that processing step is we replace attributes with particular values with other values. Here's the source code:

https://github.com/apollographql/space-kit/blob/d7663f1995f9bf86e89b129f6933c6b1ad720516/icons/src/scripts/convert.js#L60-L65

This will replace attributes (the key word being _attributes_), with something else that we can then use to style the SVGs.

Using this style will not let us do that. If we want to both be able to use this SVG in an `` tag and convert it to a colorable React component, then we'll have to change the SVG contents to suit our conversation script.

Action item: please remove the `stroke` from the css (leave the miterlimit, no idea what that is 🤷‍♂). Add `stroke="#000"` to the SVG alongside wherever that we use `class="st0"`. That will let you use the SVG as an `` and will allow it to be consumed by the conversation script so we can color it with the React component's `color`.

PR #14, opened 2019-07-01
AP-543: Create buttons

justinanastos:
TODO:

- [x] Document usage of style system

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
   - [Colors](#Colors)
   - [Icons](#Icons)
   - [Typography](#Typography)
+  - [Buttons](#Buttons)
 - [Developing Space Kit](#Developing-Space-Kit)
   - [Icons](#Icons-1)
   - [TypeScript](#TypeScript)
@@ -42,7 +43,7 @@ function MyComponent() {
       style={{
         backgroundColor: colors.indigo.dark,
         color: "white",
-        border: `1px solid ${colors.grey.light}`
+        border: `1px solid ${colors.grey.light}`,
       }}
     >
       ...
@@ -111,7 +112,7 @@ import { colors } from "@apollo/space-kit";
 const StyledButton = styled.button({
   backgroundColor: colors.indigo.dark,
   color: "white",
-  border: `1px solid ${colors.grey.light}`
+  border: `1px solid ${colors.grey.light}`,
 });
 
 function MyComponent() {
@@ -120,7 +121,7 @@ function MyComponent() {
       <StyledButton>Save</StyledButton>
       <StyledButton
         style={{
-          backgroundColor: colors.red.base
+          backgroundColor: colors.red.base,
         }}
       >
         Delete
@@ -162,13 +163,13 @@ module.exports = {
         Object.entries(typographyConfig.base).reduce(
           (accumulator, [name, properties]) => ({
             ...accumulator,
-            [`.space-kit-${name}`]: properties
+            [`.space-kit-${name}`]: properties,
           }),
           {}
         )
       );
-    }
-  ]
+    },
+  ],
 };
 ```
 
@@ -182,10 +183,10 @@ module.exports = {
     function spaceKitTypography({ addUtilities }) {
       addUtilities({
         ".space-kit-title": typographyConfig.base.xxxlarge,
-        '.caption': typographyConfig.base.xsmall
+        ".caption": typographyConfig.base.xsmall,
       });
-    }
-  ]
+    },
+  ],
 };
 ```
 
@@ -199,6 +200,67 @@ export const IconServiceItem: React.FC = () => (
   <div className="w-5 h-5">
     <IconServices className="w-full h-full text-teal" />
   </div>
+);
+```
+
+### Buttons
+
+Zeplin: https://zpl.io/amdN6Pr
+
+This is our style system for buttons. This is intended to be used to create your project's style guide (as opposed to using this component directly in your project).
+
+#### Gotchas
+
+- Apply colors with emotion or classNames.
+- If you want to override the disabled styles, you need to apply a class for `:disabled` _and_ `:disabled:hover` to make sure the disabled button doesn't have it's styles overriden by the `:hover` styles.
+
+#### Tailwind Example
+
+```js
+import React from "react";
+import classnames from "classnames";
+import { Button } from "@apollo/space-kit/Button";
+
+export const PrimaryButton: React.FC<ComponentProps<typeof Button>> = ({
+  children,
+  className,
+  ...otherProps
+}) => (
+  <Button
+    {...otherProps}
+    className={classnames(
+      className,
+      "bg-indigo hover:bg-indigo-dark text-white"
+    )}
+  >
+    {children}
+  </Button>
+);
+```
+
+#### Emotion Example
+
+```js
+/** @jsx jsx */
+import React from "react";
+import { Button } from "@apollo/space-kit/Button";
+import { colors } from "@apollo/space-kit/colors";
+import { jsx } from "@emotion/core";
+
+export const PrimaryButton: React.FC<ComponentProps<typeof Button>> = ({
+  children,
+  ...otherProps
+}) => (
+  <Button
+    {...otherProps}
+    css={{
+      backgroundColor: colors.indigo.dark,
+      color: "white",
+      hover: { backgroundColor: colors.indigo.darker },
+    }}
+  >
+    {children}
+  </Button>
 );
 ```
 
--- a/src/Button.story.tsx
+++ b/src/Button.story.tsx
@@ -142,9 +142,9 @@ storiesOf("Space Kit", module)
       >
         <VerticalButtonGroup title="Example 1">
           <Button>Rest</Button>
-          <Button forceHoverState>Hover</Button>
-          <Button forceActiveState>Active</Button>
-          <Button forceFocusState>Focused</Button>
+          <Button data-force-hover-state="true">Hover</Button>
+          <Button data-force-active-state="true">Active</Button>
+          <Button data-force-focus-state="true">Focused</Button>
           <Button disabled>Disabled</Button>
         </VerticalButtonGroup>
 
@@ -159,9 +159,9 @@ storiesOf("Space Kit", module)
           }}
         >
           <Button>Rest</Button>
-          <Button forceHoverState>Hover</Button>
-          <Button forceActiveState>Active</Button>
-          <Button forceFocusState>Focused</Button>
+          <Button data-force-hover-state="true">Hover</Button>
+          <Button data-force-active-state="true">Active</Button>
+          <Button data-force-focus-state="true">Focused</Button>
           <Button disabled>Disabled</Button>
         </VerticalButtonGroup>
 
@@ -176,9 +176,9 @@ storiesOf("Space Kit", module)
           }}
         >
           <Button>Rest</Button>
-          <Button forceHoverState>Hover</Button>
-          <Button forceActiveState>Active</Button>
-          <Button forceFocusState>Focused</Button>
+          <Button data-force-hover-state="true">Hover</Button>
+          <Button data-force-active-state="true">Active</Button>
+          <Button data-force-focus-state="true">Focused</Button>
           <Button disabled>Disabled</Button>
         </VerticalButtonGroup>
       </DemoSection>
@@ -189,9 +189,9 @@ storiesOf("Space Kit", module)
       >
         <VerticalButtonGroup title="Secondary">
           <Button>Rest</Button>
-          <Button forceHoverState>Hover</Button>
-          <Button forceActiveState>Active</Button>
-          <Button forceFocusState>Focused</Button>
+          <Button data-force-hover-state="true">Hover</Button>
+          <Button data-force-active-state="true">Active</Button>
+          <Button data-force-focus-state="true">Focused</Button>
           <Button disabled>Disabled</Button>
         </VerticalButtonGroup>
 
@@ -202,9 +202,9 @@ storiesOf("Space Kit", module)
           }}
         >
           <Button>Rest</Button>
-          <Button forceHoverState>Hover</Button>
-          <Button forceActiveState>Active</Button>
-          <Button forceFocusState>Focused</Button>
+          <Button data-force-hover-state="true">Hover</Button>
+          <Button data-force-active-state="true">Active</Button>
+          <Button data-force-focus-state="true">Focused</Button>
           <Button disabled>Disabled</Button>
         </VerticalButtonGroup>
 
@@ -221,12 +221,12 @@ storiesOf("Space Kit", module)
           <React.Fragment>
             <Button
               css={{ marginRight: 6 }}
-              forceHoverState
+              data-force-hover-state="true"
               icon={iconElement}
             />
             <Button
               css={{ marginLeft: 6 }}
-              forceHoverState
+              data-force-hover-state="true"
               fab
               icon={iconElement}
             />
@@ -234,12 +234,12 @@ storiesOf("Space Kit", module)
           <React.Fragment>
             <Button
               css={{ marginRight: 6 }}
-              forceActiveState
+              data-force-active-state="true"
               icon={iconElement}
             />
             <Button
               css={{ marginLeft: 6 }}
-              forceActiveState
+              data-force-active-state="true"
               fab
               icon={iconElement}
             />
@@ -247,12 +247,12 @@ storiesOf("Space Kit", module)
           <React.Fragment>
             <Button
               css={{ marginRight: 6 }}
-              forceFocusState
+              data-force-focus-state="true"
               icon={iconElement}
             />
             <Button
               css={{ marginLeft: 6 }}
-              forceFocusState
+              data-force-focus-state="true"
               fab
               icon={iconElement}
             />
@@ -282,9 +282,9 @@ storiesOf("Space Kit", module)
           }}
         >
           <Button>Rest</Button>
-          <Button forceHoverState>Hover</Button>
-          <Button forceActiveState>Active</Button>
-          <Button forceFocusState>Focused</Button>
+          <Button data-force-hover-state="true">Hover</Button>
+          <Button data-force-active-state="true">Active</Button>
+          <Button data-force-focus-state="true">Focused</Button>
           <Button disabled>Disabled</Button>
         </VerticalButtonGroup>
 
@@ -302,9 +302,9 @@ storiesOf("Space Kit", module)
           }}
         >
           <Button>Rest</Button>
-          <Button forceHoverState>Hover</Button>
-          <Button forceActiveState>Active</Button>
-          <Button forceFocusState>Focused</Button>
+          <Button data-force-hover-state="true">Hover</Button>
+          <Button data-force-active-state="true">Active</Button>
+          <Button data-force-focus-state="true">Focused</Button>
           <Button disabled>Disabled</Button>
         </VerticalButtonGroup>
 
@@ -322,9 +322,9 @@ storiesOf("Space Kit", module)
           }}
         >
           <Button>Rest</Button>
-          <Button forceHoverState>Hover</Button>
-          <Button forceActiveState>Active</Button>
-          <Button forceFocusState>Focused</Button>
+          <Button data-force-hover-state="true">Hover</Button>
+          <Button data-force-active-state="true">Active</Button>
+          <Button data-force-focus-state="true">Focused</Button>
           <Button disabled>Disabled</Button>
         </VerticalButtonGroup>
       </DemoSection>
--- a/src/Button.story.tsx
+++ b/src/Button.story.tsx
@@ -1,0 +1,332 @@
+/** @jsx jsx */
+import { jsx, ObjectInterpolation, css } from "@emotion/core";
+import React, { ComponentProps } from "react";
+import { storiesOf } from "@storybook/react";
+import { Button } from "./Button";
+// @ts-ignore We need to use `.tsx` because this will live next to the
+// transpiled files that we can't import.
+import { IconShip2 } from "../icons/IconShip2.tsx";
+import { colors } from ".";
+import * as typography from "./typography";
+import { DemoSection } from "./shared/DemoSection";
+import classnames from "classnames";
+
+const iconElement = <IconShip2 css={{ width: "100%", height: "100%" }} />;
+
+// This is a clever way of allowing us to use `cloneElement` while still giving
+// us the `jsx` transform for emotion.
+// @see https://github.com/emotion-js/emotion/issues/1102#issuecomment-446887725
+const cloneElement = (element: any, props: any) =>
+  jsx(element.type, {
+    key: element.key,
+    ref: element.ref,
+    ...element.props,
+    ...props,
+  });
+
+/**
+ * Wrap children with a rounded border
+ */
+const ButtonWrapper: React.FC = ({ children, ...otherProps }) => (
+  <div
+    {...otherProps}
+    css={{
+      border: `1px solid ${colors.silver.dark}`,
+      borderRadius: 8,
+      margin: 6,
+    }}
+  >
+    {children}
+  </div>
+);
+
+const VerticalButtonGroup: React.FC<{
+  buttonCss?: ObjectInterpolation<any>;
+  buttonProps?: Partial<Omit<ComponentProps<typeof Button>, "children">>;
+  children: JSX.Element | JSX.Element[];
+  darkButtonCss?: ObjectInterpolation<any>;
+  darkButtonProps?: Partial<Omit<ComponentProps<typeof Button>, "children">>;
+  title: string;
+}> = ({
+  buttonProps = {},
+  darkButtonProps = {},
+  children,
+  title,
+  buttonCss = {},
+  darkButtonCss = {},
+  ...otherProps
+}) => (
+  <div {...otherProps} css={{ margin: "0 20px", width: 300 }}>
+    <hr
+      style={{
+        height: 1,
+        borderWidth: 0,
+        backgroundColor: colors.silver.dark,
+        marginBottom: 24,
+      }}
+    />
+    <div
+      css={{
+        ...typography.base.base,
+        textTransform: "uppercase",
+        fontWeight: 600,
+        margin: 6,
+      }}
+    >
+      {title}
+    </div>
+    <div css={{ display: "flex", flexWrap: "wrap" }}>
+      <ButtonWrapper>
+        {React.Children.map(children, child => (
+          <div css={{ margin: 12 }}>
+            {cloneElement(child as any, {
+              ...buttonProps,
+              ...darkButtonProps,
+              css: buttonCss,
+            })}
+          </div>
+        ))}
+      </ButtonWrapper>
+      <ButtonWrapper
+        css={{
+          backgroundColor: colors.black.base,
+        }}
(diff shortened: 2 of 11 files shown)


src/Button.story.tsx
```diff
@@ -0,0 +1,309 @@
+/** @jsx jsx */
+import { jsx, css } from "@emotion/core";
```
mayakoneval: are we making emotion the css in js library of choice for space kit?

src/Button.story.tsx
```diff
@@ -0,0 +1,309 @@
...
+import { Button } from "./Button";
+// @ts-ignore We need to use `.tsx` because this will live next to the
+// transpiled files that we can't import.
+import { IconShip2 } from "../icons/IconShip2.tsx";
+import { colors } from ".";
+import * as typography from "./typography";
+
+const iconElement = <IconShip2 css={{ width: "100%", height: "100%" }} />;
+
+const DemoSection: React.FC<{
```
mayakoneval: This is going to be a constant throughout the space kit story books, so many we can make it a shared RFC?

src/Button.story.tsx
```diff
@@ -0,0 +1,309 @@
...
+  title: string;
+}> = ({
+  buttonProps = {},
+  darkButtonProps = {},
+  children,
+  title,
+  ...otherProps
+}) => (
+  <div {...otherProps} css={{ margin: "0 20px", width: 300 }}>
+    <hr
```
mayakoneval: Why can't we use the existing typography component to style the titles of space kit stories?

src/Button.story.tsx
justinanastos: Yes. Here's the quip doc where we discussed it: https://apollographql.quip.com/5CRRAypIxQla/Website-Development-at-Apollo

Please read the document and share any and all feedback you have! Remember, your experience is important and your opinion matters.

src/Button.story.tsx
justinanastos: That's a great idea. Yes!

src/Button.story.tsx
justinanastos: @mayakoneval 

Can you clarify what you mean? I don't understand because:

1. We don't have typography components; we only have constants defining the css properties
2. We _are_ using `typography.base` wherever we're defining typography styles

mayakoneval: oh yep, totally. tx!

mayakoneval:
Is the table POC code supposed to be in this PR as well? I see `Table.story.tsx` and `Table.tsx`

src/Button.story.tsx
```diff
@@ -0,0 +1,294 @@
...
+          </React.Fragment>
+          <React.Fragment>
+            <Button css={{ marginRight: 6 }} disabled icon={iconElement} />
+            <Button css={{ marginLeft: 6 }} disabled fab icon={iconElement} />
+          </React.Fragment>
+        </VerticalButtonGroup>
+      </DemoSection>
+
+      <DemoSection
+        title="Hidden Button"
```
timbotnik: I feel like the term `hidden` is reserved for something that shouldn't be displayed _at all_.  Might I suggest we use a different term for this style of button?  I see the word `simple` being used in the description, and much prefer that.  I think it would make sense to use a property with a string literal union prop like `feel: 'normal' | 'simple'` rather than a boolean.
 

Perhaps this dovetails with another question, which is whether we'd consider "link" styled clickable things as being replaced by this kind of button?

cc @adamatorres

src/Button.tsx
```diff
@@ -0,0 +1,291 @@
...
+  hidden?: boolean;
+
+  /**
+   * Color to make the background for a hidden button that is active.
+   *
+   * This color should be darker than the `hoverBackgroundColor`.
+   *
+   * This should not be used for a non-hidden button.
+   */
+  hiddenActiveBackgroundColor?: CSS.BackgroundColorProperty;
```
timbotnik: Along with the previous comment, I guess I'm not sure why we need a separate set of colors just for the "hidden" style?  Like can't we just provide an `activeBackgroundColor` and `focusedTextColor`?  Do we also need a `textColor` property?

src/Button.tsx
```diff
@@ -0,0 +1,291 @@
...
+ * There are way too many configuration options to use this component directly.
+ *
+ * @see https://zpl.io/amdN6Pr
+ */
+export const Button: React.FC<Props> = ({
+  backgroundColor = colors.silver.light,
+  children,
+  disabled = false,
+  disabledBackgroundColor = colors.silver.dark,
+  disabledTextColor = colors.grey.light,
```
cheapsteak: we should probably decide on whether we use styled system, or at least themeprovider (imo we should), and if we do then these would make sense in the theme context instead of ad-hoc props

README.md
```diff
@@ -202,6 +203,21 @@ export const IconServiceItem: React.FC = () => (
...
+
+This is our style system for buttons. This is intended to be used to create your project's style guide (as opposed to using this component directly in your project).
+
+#### Example
+
+```js
+import React from 'react';
+import { Button } from '@apollo/space-kit/Button';
+
+
```
justinanastos: This example could use a little more work.

src/Button.tsx
justinanastos: Really good question.

The "active" state only changes for hidden buttons, not for regular buttons.

Note that the rest state's background color matches the active state's background color:

image

Note here that the rest state's background color differs from the active state's background color:

image

src/Button.tsx
```diff
@@ -0,0 +1,291 @@
...
+  // `forceFocusState` is used in Storybook
+  const focusProperties = {
+    // The `box-shadow` property is copied directly from Zeplin
+    boxShadow:
+      "0 1px 4px 0 rgba(18, 21, 26, 0.08), 0 0 0 2px #bbdbff, inset 0 0 0 1px #2075d6, inset 0 -1px 0 0 rgba(18, 21, 26, 0.05)",
+    ...(hidden && { color: `${hiddenFocusedColor} !important` })
+  };
+
+  return (
+    <button
```
justinanastos: this needs to passthrough `disabled` to stop disabled buttons from being tab stops

src/Button.tsx
timbotnik: I guess I still don't understand why we would refuse to support an always-available  `activeBackgroundColor` and `focusedTextColor` property, feels like it could come in handy?

src/Button.tsx
```diff
@@ -0,0 +1,291 @@
...
+          : size === "large"
+          ? 112
+          : 100,
+        padding: size === "small" ? `5px 8px` : size === "large" ? 8 : 7,
+        ...(size === "small"
+          ? base.small
+          : size === "large"
+          ? base.large
+          : base.base),
+        ...(!disabled &&
```
trevorblades: Rather than changing styles based on the `disabled` prop, we can pass along `disabled` to the `button` component and target the `[disabled]` CSS selector:

```js
{
  // non-disabled styles
  '[disabled]': {
    color: 'grey'
  }
}
```

src/Button.tsx
```diff
@@ -0,0 +1,291 @@
...
+          : disabled
+          ? disabledBackgroundColor
+          : backgroundColor,
+        borderRadius: fab ? "100%" : 4,
+        borderWidth: 0,
+        ...(!hidden && {
+          boxShadow:
+            "0 1px 4px 0 rgba(18, 21, 26, 0.08), inset 0 0 0 1px rgba(18, 21, 26, 0.2), inset 0 -1px 0 0 rgba(18, 21, 26, 0.05)"
+        }),
+        minWidth: iconOnly
```
trevorblades: Instead of using these ternaries within our object definition, maybe it would be more easily understandable if we break these logical statements out to their own functions. For example:

```js
function getMinWidth(iconOnly, size) {
  switch(size) {
    case 'small':
      return iconOnly ? 28 : 76;
    case 'large':
      return iconOnly ? 42 : 112;
    default:
      return iconOnly ? 36 : 100;
  }
}

<button
  css={{
    minWidth: getMinWidth(iconOnly, size)
  }}
/>
```

src/Button.tsx
```diff
@@ -0,0 +1,291 @@
...
+            : size === "large"
+            ? 42
+            : 36
+          : size === "small"
+          ? 76
+          : size === "large"
+          ? 112
+          : 100,
+        padding: size === "small" ? `5px 8px` : size === "large" ? 8 : 7,
+        ...(size === "small"
```
trevorblades: Similar to the comment above about `minWidth`, we could do something like this here:

```js
function getFontStyles(size) {
  switch(size) {
    case 'small':
      return base.small;
    case 'large':
      return base.large;
    default:
      return base.base;
  }
}

{
  ...getFontStyles(size)
}
```

Alternatively, since the size prop lines up with the keys of the typography `base` we could do something like this:

```js
{
  ...base[size || 'base']
}
```

src/Button.tsx
```diff
@@ -0,0 +1,291 @@
...
+      "0 1px 4px 0 rgba(18, 21, 26, 0.08), 0 0 0 2px #bbdbff, inset 0 0 0 1px #2075d6, inset 0 -1px 0 0 rgba(18, 21, 26, 0.05)",
+    ...(hidden && { color: `${hiddenFocusedColor} !important` })
+  };
+
+  return (
+    <button
+      {...otherProps}
+      css={{
+        appearance: "button",
+        backgroundColor: hidden
```
trevorblades: If we define disabled styles under the `[disabled]` selector, this ternary could be simplified:

```js
{
  backgroundColor: hidden ? 'transparent' : backgroundColor,
  '[disabled]': {
    backgroundColor: disabledBackgroundColor
  }
}
```

src/Button.tsx
```diff
@@ -0,0 +1,291 @@
...
+  }
+
+  // We're storing this because it can be used for the `:active` style or when
+  // `forceActiveState` is used in Storybook
+  const activeProperties = {
+    ...(hidden
+      ? { backgroundColor: hiddenActiveBackgroundColor }
+      : {
+          // The `box-shadow` property is copied directly from Zeplin
+          boxShadow:
```
trevorblades: In the docs, I made a function for getting box shadows since some of the button states have box shadows in common. Here's how it looks:

```js
import {colors} from '@apollo/space-kit';
import {transparentize} from 'polished';

const highlightColor = 
const shadowColor = colors.black.darker;
function getBoxShadow(state) {
  const isFocused = state === 'focused';
  return [
    state === 'active'
      ? `inset 0 2px 2px 0 ${transparentize(0.88, shadowColor)}`
      : `0 1px 4px 0 ${transparentize(0.92, shadowColor)}`,
    `inset 0 -1px 0 0 ${transparentize(0.95, shadowColor)}`,
    isFocused && `0 0 0 2px ${colors.blue.lightest}`,
    `inset 0 0 0 1px ${
      isFocused ? colors.blue.lighter : transparentize(0.8, shadowColor)
    }`
  ]
    .filter(Boolean)
    .toString();
}

<button
  css={{
    boxShadow: getBoxShadow(),
    ':hover': {
      boxShadow: getBoxShadow('hover')
    },
    ':active': {
      boxShadow: getBoxShadow('active')
    },
    ':focus': {
      boxShadow: getBoxShadow('focused')
    }
  }}
/>
```

src/Button.tsx
```diff
@@ -0,0 +1,291 @@
...
+ * @see https://zpl.io/amdN6Pr
+ */
+export const Button: React.FC<Props> = ({
+  backgroundColor = colors.silver.light,
+  children,
+  disabled = false,
+  disabledBackgroundColor = colors.silver.dark,
+  disabledTextColor = colors.grey.light,
+  fab = false,
+  forceActiveState = false,
```
trevorblades: We should consider if we want to be exposing the different button states in Storybook, or if displaying a single interactive button for each variant and size is enough. It might be nice to not need to support these `force*` props

src/Button.story.tsx
justinanastos: Two things:

1. Love the `feel` or `variant` prop instead of a boolean `'hidden'`.
2. A clickable link; if we want it to have the same padding/spacing, then yes 🤷‍♂

src/Button.tsx
justinanastos: You're totally right @timbotnik. We changed direction to have sensible defaults for the colors, but all other configuration for colors will have to be with either tailwind classes or emotion styles instead of props. That will reduce the surface area of the API dramatically.

cheapsteak:
Will take a look at the rest of this first thing Monday  

The thing that stuck out to me right now though is - can we not use object-based styles? `camelCasedFields` and `"quoted-kebab-values"` is the kind of annoyance that first-generation css-in-js had to suffer. 

btw we should individually install https://marketplace.visualstudio.com/items?itemName=jpoissonnier.vscode-styled-components (and add it to recommended plugins) so we get syntax highlighting and css validation and auto-completion for

```
css`css strings`
```

src/Button.tsx
```diff
@@ -0,0 +1,241 @@
...
+      css={{
+        "&[disabled]": {
+          backgroundColor: colors.silver.dark,
+          color: colors.grey.light,
+
+          // We need to also set the `:hover` on `:disabled` so it has a higher
+          // specificity than any `:hover` classes passed in. This also means
+          // that both of these need to be overriden if we want to use a custom
+          // disabled color.
+          ":hover": {
```
trevorblades: Alternatively, we could define non-disabled hover styles like this:

```js
{
  ':not([disabled]):hover': { // this selector might need a & before :not
    // hover styles here
  }
}
```

justinanastos: I wanted to avoid complicated css Nesting here because it would be impossible to represent overrides for this with tailwindcss, which would make this library less extensible.

src/Button.tsx
cheapsteak: A bit surprised to see these. I recall when we discussed this over the kitchen table, you were a bit perplexed about why I was considering setting active state through props as part of the button's API, and wanted to forego that until we needed this in any project

I'm guessing then that these are solely for the purpose of being able to show them in storybook. I agree that there is value in having that to allow  visual regression tests. I don't think adding public props solely for storybook-ability is a good idea. E.g. when I'm using vscode's code completion to see what a components props are, I don't want to see props that I'm not supposed to use

Let's use class names here instead. We should be able to achieve the same result without exposing test-apis as props

src/Button.tsx
justinanastos: Fantastic. Done.

src/Button.tsx
justinanastos: I'm going to leave this as it is now because I think it boils down to personal preference and could end up in bikeshedding. If we find need to revisit this because we have problems, then let's revisit it.

src/Button.tsx
justinanastos: I would prefer to be able to copy-paste the values directly from the zeplin designs instead of doing this. Thanks for diving deep into this :)

src/Button.tsx
```diff
@@ -0,0 +1,241 @@
...
+    React.ButtonHTMLAttributes<HTMLButtonElement>,
+    HTMLButtonElement
+  > {
+  /**
+   * Treat this button as a floating action button
+   *
+   * You must include an `icon` prop and you must _not_ include a `children`
+   * prop for a floating action button.
+   */
+  fab?: boolean;
```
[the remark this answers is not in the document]
justinanastos: Good question @cheapsteak . `fab` and `variant` are orthogonal. I prefer sticking with FAB because (with an obviously not statistically significant sample size of 2), I know what it means and that's what @adamatorres used in the Zeplin doc. FAB = float action button. That can be of the normal or simple variant.

image

src/Button.tsx
```diff
@@ -0,0 +1,241 @@
...
+  /**
+   * Which variant to display
+   *
+   * The options are as follows:
+   *
+   * - `normal`: A button with a border and a background
+   * - `simple`: No border or background (overrideable).
+   *     - This will require you adding an `:active` style in addition to `:hover`
+   */
+  variant?: "normal" | "simple";
```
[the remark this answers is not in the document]
justinanastos: It previously was a boolean flag; @timbotnik 	suggested we remove the boolean flag and instead use something more descriptive. I'm very happy to use something besides "simple" and "normal". I don't want to bikeshed this though.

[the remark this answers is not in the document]
cheapsteak: I found the comment, would you mind writing a bit about the reasoning behind it being more sensible to have a string enum here?  

I agree that `hidden` shouldn't be used since that's appropriating something else, but switching away from a boolean prop makes usage more verbose (`<Button simple/>` vs `<Button variant="basic"/>`), and also makes condition-checks in the component noisier (`props.simple ? ... : ...` vs `props.variant === 'simple' ? ... : ...`). What are these tradeoffs in service of?

src/Button.tsx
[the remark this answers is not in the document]
cheapsteak: A "Floating Action Button" at least to me surely _sounds_ like a variant of a "Button", and that it's orthogonal to the variant prop is imo a description of the outcome of API design (implementation detail) and imo not necessarily what makes intuitive sense  

Using boolean props instead of a `variant` enum prop would allow us to avoid the question of 'why is X a "variant" and Y not a "variant"?'  

Consumers (or space-kit itself) could make a `<Fab/>` component that's just `(...props) => <Button {...props} circular/>`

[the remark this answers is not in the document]
cheapsteak: Does it make sense that I think of "small"/"large"/"normal" as specifying a "size" variant of a Button, and "fab"/"circular" as specifying a "shape" variant?

src/Button.tsx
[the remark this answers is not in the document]
trevorblades: How about `raised` and `flat`?

src/Button.tsx
```diff
@@ -0,0 +1,192 @@
...
+            : size === "large"
+            ? base.large
+            : base.base),
+
+          fontWeight: 600,
+          // Disable the outline because we're setting a custom `:active` style
+          outline: 0,
+        },
+
+        // This must come after `:focus` or the `:focus` state will override `:active`
```
trevorblades: We could write this like this to add more CSS specificity and avoid the condition mentioned in this comment:

```js
{
  '&:not([disabled])': {
    ':hover': ...
  }
}
```

src/Button.tsx
```diff
@@ -0,0 +1,192 @@
...
+        !disabled && {
+          ":hover, &[data-force-hover-state]": {
+            backgroundColor: colors.silver.base,
+            cursor: "pointer",
+            ...(variant === "normal" && {
+              // The `box-shadow` property is copied directly from Zeplin
+              boxShadow:
```
trevorblades: I found the `...(condition && { ... })` syntax difficult to understand here. What if we used a ternary alternating `boxShadow` between a valid value and `none`?

```js
boxShadow:
  variant === 'normal'
    ? '0 5px 10px 0 rgba(18, 21, 26, 0.12), inset 0 0 0 1px rgba(18, 21, 26, 0.2), inset 0 -1px 0 0 rgba(18, 21, 26, 0.05)' 
    : 'none'
```

Or better yet...

```js
const normalHoverShadow = [
  '0 5px 10px 0 rgba(18, 21, 26, 0.12)',
  'inset 0 0 0 1px rgba(18, 21, 26, 0.2)',
  'inset 0 -1px 0 0 rgba(18, 21, 26, 0.05)'
];

boxShadow: variant === 'normal' ? normalHoverShadow.toString() : 'none'
```

src/Button.tsx
```diff
@@ -0,0 +1,192 @@
...
+            }
+          }
+        },
+
+        variant === "simple" && { backgroundColor: "transparent" },
+        {
+          borderRadius: fab ? "100%" : 4,
+
+          borderWidth: 0,
+          ...(variant === "normal" && {
```
trevorblades: Same as https://github.com/apollographql/space-kit/pull/14/files#r305962684

src/Button.tsx
```diff
@@ -0,0 +1,192 @@
...
+            // that both of these need to be overriden if we want to use a custom
+            // disabled color.
+            ":hover": {
+              backgroundColor: colors.silver.dark,
+              color: colors.grey.light
+            }
+          }
+        },
+
+        variant === "simple" && { backgroundColor: "transparent" },
```
trevorblades: Should we be setting a default background color for non-simple buttons?

```js
backgroundColor: variant === 'simple' ? 'transparent' : colors.silver.light
```

src/Button.tsx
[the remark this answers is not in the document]
timbotnik: @trevorblades I like `raised` and `flat`!  The general reasoning behind not using a boolean is so that you can't provide an API that allows contradictions, i.e. `<Button raised flat>`.  But perhaps we'd want to deal with this via PropType checking at run time?  I'm not going to hold my position that firmly, it's not a very big deal to me in the end, so also happy to be the odd man out.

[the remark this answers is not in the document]
cheapsteak: I don't think we'd need both props, just one should be enough?   

`<Button/>` would be the raised/normal button
`<Button flat/>` would be flat

src/Button.tsx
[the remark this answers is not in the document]
justinanastos: I think this is getting into bikeshed territory. We're talking about different words to describe the same things.

I don't think the fab prop is going to cause an issue; if you think this is a blocker for the PR; then let's jump on a call. Otherwise, let's just leave it.

src/Button.tsx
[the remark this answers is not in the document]
justinanastos: I will hold strong on not allowing contradicting props; especially where we don't have _any_ dynamic propType checking.

I accidentally used `variant`, and I should have used `feel` as @timbotnik 	originally suggested. I'm totally cool with `raised` and `flat`, as long as they are string props passed somewhere.

I read an article a few weeks ago: [How to Avoid the “Boolean Trap” When Designing React Components](https://spicefactory.co/blog/2019/03/26/how-to-avoid-the-boolean-trap-when-designing-react-components/) and agree that using an enum for the different options feels more correct.

Yes, it's more verbose; I don't think that's a bad thing. It is still readable, the different enum options can be documented in the `interface` so you get helpful type hinting with TS.

src/Button.tsx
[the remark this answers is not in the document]
justinanastos: I think that `variant` makes sense for this. I can make a `variant` prop that takes `'circular' | undefined`

src/Button.tsx
justinanastos: I mentioned this before; we this will make overrides with tailwind much more difficult because of that increased specificity.

src/Button.tsx
justinanastos: I think understanding what the `...()` means is important. That will either spread the values in the `{}` into the object or add nothing to the object.

justinanastos: I also think this is adding more code and an unnecessary default back to the css default instead of doing the right thing: adding nothing in certain circumstances.

src/Button.tsx
justinanastos: Yes we should! Good catch @trevorblades

src/Button.tsx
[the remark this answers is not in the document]
cheapsteak: There is no contradicting prop  

`flat?: boolean` and that's it. If it's true then it's flat, if it's false then it's raised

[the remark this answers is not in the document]
trevorblades: Yea, that makes sense to me. Considering we only have two button variants to represent, we could use a single `flat` boolean. `variant` allows us the ability to add additional variants in the future without changing the API, but we also may not want to worry about future changes/additions at this point.

@timbotnik we're on the same page with regards to the two-boolean thing. 👍  I was suggesting we use `flat` and `raised` as values for the `variant` prop.

src/Button.tsx
trevorblades: Oh yea, I forgot we had talked about that!

src/Button.tsx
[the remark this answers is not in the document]
cheapsteak: I'm not advocating for `fab` or `size` to be renamed to "variant", my point is that a "variant" prop for a `<Button/>` suffers from a similar imprecision as the word "account" does in (organizations have accounts, users have accounts, which type of account "account" refers to is ambiguous)   

To escape ambiguity/imprecision, either no prop should be called "variant" (because every prop is a variant), or props should make explicit what they are variating (`userAccount`, `orgAccount`; `sizeVariant`, `shapeVariant`)

src/Button.tsx
trevorblades: For me, this is just about representing the conditional `boxShadow` style in a way that makes sense at first glance. If hardcoding `'none'` is not desirable, you could likely do this as well:

```js
boxShadow: variant === 'normal' && 'box shadow string'
```

If I remember correctly, Emotion will not include a box shadow property if the value is falsey.

src/Button.tsx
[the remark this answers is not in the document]
justinanastos: I think we're coming down to personal preferences and playing devil's advocate now. I'd prefer to leave this as a `feel?: 'raised' | 'flag'`.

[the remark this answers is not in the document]
cheapsteak: Specifically regarding the article; Uncle Bob is indeed frequently relevant, but I believe the author of the post is mistaken in using it for his argument here  

"loudly proclaiming that this function does more than one thing" is not a problem that enums solve (if anything, a boolean flag proclaims to do only two things, while an enum proclaims to do an unknown and likely higher number (until you mouse over to see exactly how many)). The problem exists in the scope that a function (/component) has chosen to encapsulate. At the end of the "Flag Arguments" section, he writes "We should have split the function into two: renderForSuite() and renderForSingleTest()." -- the solution to flag arguments in his opinion is to split into separate functions. This is one of those sections that imo might not translate as directly to UI development (unless we choose to split this one component into separate `<SimpleButton/>` and `<Button/>` components)

Another reason why he's so against flag arguments is in the next paragraph of the quote from the article - "[..] the method call `render(true)` is just plain confusing to a poor reader. Mousing over the call and seeing `render(boolean isSuite)` helps a little, but not that much. [..]". boolean props in React don't suffer from the same confusion.

Button.d.ts
```diff
@@ -0,0 +1,23 @@
+import React from "react";
```
cheapsteak: I don't think we want to check this in do we?

Button.js
```diff
@@ -0,0 +1,73 @@
+"use strict";
```
cheapsteak: Same for this and Button.js.map, let's not check in compiled code

src/Button.tsx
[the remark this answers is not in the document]
justinanastos: I think we've reached a concesnsus that what I've got here is ok and I wanted to leave one more comment after a good night's sleep and after reading reading https://overreacted.io/optimized-for-change/ again.

I want to avoid boolean props because there's a possibility that we'll come up with more button variants than "fab" in the future, so using an optional string prop with the only option being `"fab"` makes those changes backwards compatible. I weighed this decision against "solve today's problems" and decided this solves today's problems well and is optimized for change.

src/Button.tsx
justinanastos: I understand the logic behind defining an array of the shadows and then using `toString()` on it; however that defeats the goal of copy-pasting the values in directly from zeplin. I opted to use the `? ... : 'none'` because I needed to to fix weird functionality glitches.

I'm going to leave this with the new update now 👍

Button.d.ts
justinanastos: Good catch @cheapsteak . Thank you.

PR #16, opened 2019-07-02
AP-546: Tables POC

justinanastos:
POC on an abstraction we could use for tables. Not sure if this is actually _better_ than just doing it by hand, but this would allow us to configure how tables look with props applied to the `Table` component 🤷‍♂


src/Table.tsx
```diff
@@ -0,0 +1,133 @@
...
+     * ```js
+     * renderTd: ({ children, ...otherProps}) => (
+     *   <td
+     *     {...otherProps}
+     *     onClick={() => alert('clicked')}
+     *   />
+     * )
+     * ```
+     */
+    renderTd?: React.FC<
```
justinanastos: I think this is the wrong way to go. I think @cheapsteak 's implementation using `tdProps` is far better.

mayakoneval: @cheapsteak do you know what tdProps justin is referring to? I would like to merge this POC or improve it and then merge so we can consume it as a beta in the space-kit npm package and use it for the new accounts tables.

mayakoneval: if not, can I release this as a beta?

cheapsteak: I agree that this `renderTd` doesn't quite feel like the right API

`tdProps` would be an alternative to `renderTd`  (can reference react-table's `getTdProps`, the important piece is the arguments it passes to the callback `(state, rowInfo, column, instance)`)   

We don't necessarily need to implement that here until we need it, but we should probably remove the `renderTd` prop and go back to what it was before the force-push, so if we do add `getTdProps` it'd only be an addition to the component's API rather than a breaking change

mayakoneval: I'm going to assume we can handle extra props with hooks anchored to the child elements and remove renderTd

cheapsteak: Don't quite follow "handle extra props with hooks anchored to the child elements", but removing `renderTd` sounds good 👍

PR #18, opened 2019-07-18
Partial proposed changes to AP-543: Create buttons

cheapsteak:
This PR removes the `forceActiveState`, `forceHoverState`, and `forceFocusState` props from `<Button/>` 

There is value in being able to snapshot different visual states for regression snapshots, this switches to directly use and require the data attributes @justinanastos already uses in the component (set via the removed props). The difference is we no-longer expose the props publicly, i.e. no auto-completion when using the component (which imo is as it should be, I wouldn't want to see props I shouldn't use in the auto-completion list)

Another minor change is to take advantage of how the `css` prop accepts arrays, so we don't have to spread conditional styles into one object (emotion handles merging for us in the expected order (same as Object.assign)), less `...` syntax noise    

I think there still could be value in exposing a boolean `active` prop (out of scope of the initial component), but that's probably the only one where it makes sense to publicly force a state   

cc @timbotnik @trevorblades @mayakoneval

### DIFF
--- a/src/Button.story.tsx
+++ b/src/Button.story.tsx
@@ -142,9 +142,9 @@ storiesOf("Space Kit", module)
       >
         <VerticalButtonGroup title="Example 1">
           <Button>Rest</Button>
-          <Button forceHoverState>Hover</Button>
-          <Button forceActiveState>Active</Button>
-          <Button forceFocusState>Focused</Button>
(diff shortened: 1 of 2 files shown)


justinanastos:
This is awesome @cheapsteak. Merging.

PR #21, opened 2019-07-19
Edit to Table Component

mayakoneval:
- no console errors about keys
- allow header titles to be react components
- no bottom border on last row
- default to full width


src/Table.tsx
```diff
@@ -71,14 +73,16 @@ export function Table<RowShape>({
       </thead>
       <tbody>
         {data.map((item, index) => (
-          <tr>
+          <tr key={index}>
```
cheapsteak: Ah good idea adding keys, we do need this  

We shouldn't use the positional index as the `key` though (this is actually the same as not specifying the prop at all)  

Say we have 3 rows (A, B, C), and we were to delete A to result in 2 rows of (B, C), we'll end up rendering (A, B) because according to the keys, `3` was removed, `1` and `2` haven't changed. ([a longer article on this](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318))

One solution would be to have `<Table/>` accept a `keyOn` prop (analogous to react-table's `indexKey` prop), then we can `<tr key={item[props.keyOn]}/>`

src/Table.tsx
```diff
@@ -71,14 +73,16 @@ export function Table<RowShape>({
...
             {columns.map(
               ({
                 render
-              }) => (
+              }, rowIndex) => (
                 <td 
+                  key={rowIndex}
```
cheapsteak: Same issue here, the `key` shouldn't be the positional index (this index also isn't the `rowIndex` but is rather the `columnIndex`)   

We should require that each `column` specify its own `id` (in addition to `headerTitle` and `render` (the ambiguity of this `render` prop might come back to bite us later if we want to differentiate between render functions for cells and headers))

src/Table.tsx
mayakoneval: @cheapsteak typescript wont allow us to access item[props.keyOn] unless keyOn is a single type of RowShape. I don't know how to type something as "one of the fields of a type", I know Partial exists but you can't access an element on a Partial typed key. Thoughts?

cheapsteak: How's this? 

```diff
@@ -36,14 +36,18 @@ interface Props<RowShape> {
       list: ReadonlyArray<RowShape>
     ) => React.ReactNode;
   }>;
+
+  keyOn: string | ((row: RowShape) => string | number);
 }
 
 export function Table<RowShape>({
   data,
   density = "standard",
   columns,
+  keyOn,
 }: Props<RowShape>): ReturnType<React.FC> {
   const padding = density === "standard" ? 8 : density === "condensed" ? 3 : 11;
+  const getRowKey = typeof keyOn === 'function' ? keyOn : (row) => row[keyOn];
 
   return (
     <table
@@ -77,7 +81,7 @@ export function Table<RowShape>({
       </thead>
       <tbody>
         {data.map((item, index) => (
-          <tr key={index}>
+          <tr key={getRowKey(item)}>
             {columns.map(
               ({
                 render,

```

(dont see any ts errors in my editor at least)

mayakoneval: oo nice! thanks

justinanastos: My advice would be to go even more extreme: provide a function to get the key and if that function doesn't exist, then use `JSON.stringify` on the whole row.

src/index.ts
```diff
@@ -1,5 +1,7 @@
-import * as colors from "./colors";
-import * as fonts from "./fonts";
+import * as colors from './colors';
+import * as fonts from './fonts';
+import {Table} from './Table';
 
-export { colors };
-export { fonts };
+export {colors};
```
justinanastos: I was being lazy not adding any CI to this project. If we're going to start having whitespace and quote style changes, it's time for CI and a prettier config.

PR #22, opened 2019-07-22
[WIP] AP-545: Modals

mayakoneval:
Modals in Space Kit !

[Design (Space Kit)
](https://app.zeplin.io/project/5c7dcb5ab4e654bca8cde54d/screen/5d0ad14c42474a710911cccd)
[Design (Engine)
](https://app.zeplin.io/project/5b3665ece41d8c86a0f30286/screen/5d0954813916305e0af0542f)
This is a very simple implementation, storybook contains 3 static modal pages and one interactive example

TODO: integrate buttons when they are ready


src/Modal.tsx
```diff
@@ -0,0 +1,114 @@
...
+              width: getModalWidth(size),
+              opacity: 1,
+              backgroundColor: "white",
+              padding: size === "large" ? "40px" : "32px",
+              borderRadius: 12,
+              position: "absolute",
+              left: "50%",
+              top: "20%",
+              transform: "translate(-50%)",
+              minWidth: 400,
```
cheapsteak: let's set a maxWidth too so content doesn't go off screen (maybe `calc(100% - 30px)` to leave some padding around the edges)

src/Modal.tsx
```diff
@@ -0,0 +1,114 @@
...
+          <div onClick={handleChildClick} css={{
+              width: getModalWidth(size),
+              opacity: 1,
+              backgroundColor: "white",
+              padding: size === "large" ? "40px" : "32px",
+              borderRadius: 12,
+              position: "absolute",
+              left: "50%",
+              top: "20%",
+              transform: "translate(-50%)",
```
cheapsteak: oldschool! nothing wrong with that :P

src/Modal.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+const getTitleTypography = ( size: Props["size"]) => {
+    if(size === "small") {
+        return {...typography.base.large,
+                fontWeight: 600};
+    }
+    return {...typography.base.xxlarge};
+}
+
+const getModalWidth = ( size: Props["size"]) => {
```
cheapsteak: (non-blocking)  

Curious how you feel about

```tsx
const modalWidths = {
  small: 460,
  standard: 600,
  large: 800,
}

// ...

modalWidths[props.size]

```

justinanastos:
```
modalWidths: Record<React.ComponentProps<Modal>["size"], CSS.WidthProperty> = {
  small: 460,
  standard: 600,
  large: 800,
}
```

I ended up changing this to use a `switch` statement with the `default` of `assertUnreachable(size)`, in which `size` is `:never`, but I think your way is way simpler.

PR #23, opened 2019-07-23
AP-543 Allow custom emotion css injection point

justinanastos:
Emotion injects `<style>` elements at the end of your page's `<head>` at runtime. This can cause those styles to be placed after the page's `.css` assets, resulting in the `<style>` winning the tie-breaker of specificity. This will prevent consumers from overriding css properties with tailwind classes.

This PR adds the ability for users to specify where on the page the styles will be injected with a specific element that can be injected into the page.

Relates to [AP-543](https://golinks.io/j/AP-543)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ tsconfig.tsbuildinfo
 
 # Typography
 /typography
+
+# EmotionWrapper
+/EmotionWrapper.*
--- a/README.md
+++ b/README.md
@@ -7,21 +7,22 @@
 
 ## Table of Contents <!-- omit in toc -->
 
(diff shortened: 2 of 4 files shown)


trevorblades:
Great work on this @justinanastos @cheapsteak! Just to be sure, if I want to opt in to this feature, would I add this to my HTML `head`?

```html
<style id="emotionStyleContainer">
```

cheapsteak:
ack @justinanastos	sorry I missed a thing   

I didn't realize the wrapper was being placed outside the Button component. That'd be like wrapping an ApolloProvider in each `<Query/>`

I though the EmotionWrapper was going to be used inside engine

PR #24, opened 2019-07-24
remove EmotionWrapper from Button

cheapsteak:
Follow up to https://github.com/apollographql/space-kit/pull/23   
I didn't realize in that PR the `EmotionWrapper` which is essentially an alias around `CacheProvider` in the PR was added to wrap around Button. `CacheProvider` isn't meant to wrap around individual components (that'd be like wrapping an ApolloProvider around each Query component)

The only thing this PR does is to remove the `<EmotionWrapper>` wrapping around the contents of Button

Projects that need to specify where styles need to be injected (afaik just Engine for now) should either import EmotionWrapper (or CacheProvider) and use it next ApolloProvider (inside/outside doesn't quite matter)

### DIFF
--- a/src/Button.tsx
+++ b/src/Button.tsx
@@ -2,7 +2,6 @@
 import * as colors from "./colors";
 import { base } from "./typography";
 import { jsx } from "@emotion/core";
-import { EmotionWrapper } from "./EmotionWrapper";
 
 // Types that could use some improvement:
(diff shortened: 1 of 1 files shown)


justinanastos:
This was merged without updating the documentation on how to use this in a project that needed the wrapper.

justinanastos:
Also no update to the changelog

PR #26, opened 2019-07-25
AP-543 Refactor button implementation

justinanastos:
There are a lot of subtlties involved in the colors and behavior of the
button. This allows us to pass a few configuration options in, including a
`color` prop, and all these color properties will be handled automatically.

### DIFF
--- a/src/Button.tsx
+++ b/src/Button.tsx
@@ -265,9 +265,7 @@ export const Button: React.FC<Props> = ({
                   : colors.grey.light,
             },
           },
-        },
 
-        {
           backgroundColor:
             feel === "raised"
               ? color
--- a/src/Button.tsx
+++ b/src/Button.tsx
@@ -215,12 +215,6 @@ export const Button: React.FC<Props> = ({
     }
   }
 
-  /**
-   * Define the padding here so we can use it for both padding-right and
-   * padding-left
-   */
-  const paddingX = size === "small" ? 8 : size === "large" ? 8 : 7;
-
   return (
     <button
       {...otherProps}
@@ -285,6 +279,10 @@ export const Button: React.FC<Props> = ({
 
           color: getTextColor({ color, feel, theme }),
 
+          // Vertically center children
+          display: "inline-flex",
+          justifyContent: "center",
+
           height: getHeight({ size }),
 
           minWidth: iconOnly
@@ -299,8 +297,9 @@ export const Button: React.FC<Props> = ({
             ? 112
             : 100,
 
-          paddingLeft: paddingX,
-          paddingRight: paddingX,
+          // We have to set the Y padding because browsers (at least Chrome) has
+          // a non-symmetrical vertical padding applied by default.
+          padding: `0 ${size === "small" ? 8 : size === "large" ? 8 : 7}px`,
 
           ...(size === "small"
             ? base.small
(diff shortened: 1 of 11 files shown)


src/Button.tsx
```diff
@@ -133,7 +301,8 @@ export const Button: React.FC<Props> = ({
             ? 112
             : 100,
 
-          padding: size === "small" ? `5px 8px` : size === "large" ? 8 : 7,
+          paddingLeft: paddingY,
```
trevorblades: Shouldn't these be called `paddingX` since they're on the left and right side?

src/Button.tsx
```diff
@@ -88,39 +215,80 @@ export const Button: React.FC<Props> = ({
...
+              color:
+                feel === "flat" && theme === "dark"
+                  ? colors.grey.dark
+                  : colors.grey.light,
             },
           },
         },
 
         {
           backgroundColor:
```
trevorblades: This object and the one above it in the array could be merged together:

```js
[
  {'&[disabled]': ...},
  {backgroundColor: ...},
  // ...other stuff
]
// can be...
[
  {
    '&[disabled]': ...,
    backgroundColor: ....,
    // ...other keys
  },
  // ...other objects
]
```

trevorblades:
Looks good @justinanastos! Just a few small comments above 👍

src/Button.tsx
justinanastos: Haha yes 🤕

mayakoneval:
I still am unable to override the text color of these buttons unless I wrap the text in its own text color, why is the default black for a red button? To fix, could we add a classNames prop?

justinanastos:
@mayakoneval 

> I still am unable to override the text color of these buttons unless I wrap the text in its own text color, why is the default black for a red button? To fix, could we add a classNames prop?

You can't override _with tailwind_ yet because of emotion. There's a way to get past that but I haven't figured it out out. We don't need to add a `className` prop because all props that aren't destructured are passed through to to the `<button>`, including `className`. In the meantime cheat: use `style={{ color: colors.white }}`.

src/Button.tsx
justinanastos: @cheapsteak explicitly separated them out, but I don't remember why. I think I'll leave it for now. This whole thing could use a clean-up, but it should be ready to ship now.

Ship when it's useful, not perfect!

cheapsteak: This is the file from when I separate them  
https://github.com/apollographql/space-kit/blob/f2ed4b77f16d6412a9ec0dc0ce5d01abdea45dd2/src/Button.tsx

At the time there was a conditional `variant === "simple" && { backgroundColor: "transparent" },` separating the 1st block and the 3rd block. 

https://github.com/apollographql/space-kit/blob/f2ed4b77f16d6412a9ec0dc0ce5d01abdea45dd2/src/Button.tsx#L105-L108

It looks like that's been removed so this should be good to merge into one object

adamatorres:
@justinanastos -- Here's what I'm seeing needs to change:
- [ ] basic simple button and simple button with icon and both light and dark versions, the focus state text color should be blue.
- [ ]  The examples for the default and small buttons, the alignment of the text is off vertical center, but like 1 or 2 pixels.

src/Button.tsx
justinanastos: Thanks @cheapsteak 	! Always good to get context from the source. Since this hasn't been approved by anyone I'll make the change.

>non blocking of course

Very much appreciate when people write this.

justinanastos:
@adamatorres great eyes. the icons and text alignment were also off in the simple buttons. Should all be correct now!

PR #29, opened 2019-07-26
RFC: Set up CI

justinanastos:
This sets up CI to check our TypeScript and eslint. Also, this adds eslint.

I tried to make the eslint configuration as not-annoying and not-opinionated as possible. It extends:

* [`eslint-config-problems`](https://github.com/RyanZim/eslint-config-problems)

    Intended to only lint things that cause errors. It doesn't apply any stylistic preference rules. Write code however you want.

 * [`eslint-plugin-react/recommended`](https://github.com/yannickcr/eslint-plugin-react)

    Just the recommended rules. I disabled a few that we weren't following; `prop-types` and `display-name` for example. We use TypeScript 100% here; we don't need propTypes.

* [`@typescript-eslint/eslint-plugin/recommended`](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin)

    Recommended TS rules with some disabled.

* [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier) and [`eslint-plugin-prettier`](https://github.com/prettier/eslint-plugin-prettier)

    Disables all style rules from all other configs for things prettier handles and adds an autofixable prettier eslint rule.

* [`eslint-plugin-react-hooks`](https://www.npmjs.com/package/eslint-plugin-react-hooks)

    React hooks rules and exhaustive dependency enforcement

Other weird things:
* I cache `node_modules` with a key based on `package-lock.json` and `.circleci/config.yml` so that we don't have to run installs. I saw it takes ~5s to restore the cache versus ~25s to run `npm ci`, so I stuck with it
* I don't cache the eslint rules between builds because it's hard to automatically invalidate the cache when rules change. Just don't bother.


justinanastos:
This touches a lot of files so I don't want to do work without having this merged because of merge conflicts. I'm going to merge this now and hope that we still have a conversation if there are any objections!

@mayakoneval @trevorblades @cheapsteak @daniman @timbotnik

PR #31, opened 2019-07-26
Move all the files to live under paths

justinanastos:
I also made the potentially controversial call to name things `Button/Button.tsx` with an `index.ts` that exports everything from `Button.tsx` so I could easily find my files with `Button.tsx`.


justinanastos:
> This seems okay to me. We are now able to import like so?
> 
> ```
> import { Button } from "@apollo/space-kit"
> ```
> 
> is that correct?

@mayakoneval This does two things:

1. You can't import anything from `@apollo/space-kit`
2. Everything lives in a folder

    This is in contrast to some things in `src` living in folders and others just being files. Now everything is in a folder.

Also; I re-requested a review because I rebased this and fixed what was broken.

PR #32, opened 2019-07-29
[AP-539] Table Additions

fmvfelipemonsalve:
[AP-539](https://golinks.io/j/AP-539)
This adds ability to set column widths for tables, plus added the loading state as spacified in the design spec. I made this modifications so I could finish AP-539.

### DIFF
--- a/src/Table/Table.story.tsx
+++ b/src/Table/Table.story.tsx
@@ -101,3 +101,67 @@ storiesOf("Table", module)
       ]}
     />
   ));
+
+storiesOf("Table", module)
+  .addDecorator(withKnobs)
+  .add("User Example Set Column Widths", () => (
+    <Table<User>
+      keyOn="name"
+      css={{ color: colors.black.base }}
+      density={
+        select("Density", ["standard", "condensed", "relaxed"], "standard") as
+          | "standard"
+          | "condensed"
+          | "relaxed"
+      }
+      // data is an array of type user (defined as a generic)
+      data={users}
+      columns={[
+        {
+          id: 1,
+          colProps: {
+            width: "10%",
+          },
+          render: ({ image }) => (
+            <img css={{ width: 32, height: 32 }} src={image} />
+          ),
+        },
+        {
+          id: 2,
+          headerTitle: "description",
+          colProps: {
+            width: "30%",
+          },
+          render: ({ name, email }) => (
+            <React.Fragment>
+              <div>{name}</div>
+              <div
+                css={{
+                  color: colors.grey.base,
+                  ...typography.base.small,
+                }}
+              >
+                {email}
+              </div>
+            </React.Fragment>
+          ),
+        },
+        {
+          id: 3,
+          headerTitle: (
+            <div
+              css={{
+                color: colors.blue.base,
+              }}
+            >
+              Date Added
+            </div>
+          ),
+          colProps: {
+            width: "60%",
+          },
+          render: ({ dateAdded }) => dateAdded,
+        },
+      ]}
+    />
+  ));
--- a/src/Table/Table.story.tsx
+++ b/src/Table/Table.story.tsx
@@ -5,7 +5,7 @@ import { storiesOf } from "@storybook/react";
 import { colors } from "../colors";
 import * as typography from "../typography";
 import { Table } from "./Table";
-import { withKnobs, select } from "@storybook/addon-knobs/react";
+import { withKnobs, select, number } from "@storybook/addon-knobs/react";
 
 interface User {
   name: string;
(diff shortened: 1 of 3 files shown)


mayakoneval:
@fmvfelipemonsalve do you mind if I add a commit to delete the left padding on the first column and the right padding on the last child? I noticed this is inconsistent with the designs we have for tables and I would like to slip it in here

fmvfelipemonsalve:
@mayakoneval Go for it. I did notice that, but didn't think it was important. I might need to make slight adjustments to the service tag tables after this, so let me know when you do this.

src/Table.tsx
```diff
@@ -12,6 +12,8 @@ interface Props<RowShape> {
 
   density?: "standard" | "condensed" | "relaxed";
 
+  loading?: boolean;
```
justinanastos: I don't think we should put this into space kit right now. First, I think the current tables implementation would allow you to show the exact same loading state as you have here because the content that's output is totally configurable, meaning you can configure it to show this loading state.

I also would like feedback from @adamatorres regarding this loading state. I'd be happy to settle for a loading spinner showing before we see this table at all like we do in other parts of the application instead of coming up with a unique loader that we're going to remove later.

justinanastos: I don't think we should put this into space kit right now. First, I think the current tables implementation would allow you to show the exact same loading state as you have here because the content that's output is totally configurable, meaning you can configure it to show this loading state.

I also would like feedback from @adamatorres regarding this loading state. I'd be happy to settle for a loading spinner showing before we see this table at all like we do in other parts of the application instead of coming up with a unique loader that we're going to remove later.

src/Table.tsx
```diff
@@ -24,6 +26,10 @@ interface Props<RowShape> {
      * an id for the column
      */
     id: string | number;
+    /**
+     * the width for the column
+     */
+    width?: string | number;
```
justinanastos: I think we can come up with a better long-term solution here. After reading [Optimized for Change](https://overreacted.io/optimized-for-change/) I've tried to think of solutions in terms of how easy to maintain they'll be in the future. Here we'd be adding a custom prop for the `column` config. What if we want to add other styles? What if we want to add other props? How can we solve those problems right now?

I think a better solution would take from the solution that @cheapsteak came up with in engine-frontend:

https://github.com/mdg-private/engine-frontend/blob/527b6b9d644fe7150e73e6a1f49c1516048545fb/imports/ui/components/TableWithDrawers.tsx#L37-L41

I think we should create an option in `column` for that will take either an object or a function that accepts `RowData` and returns an object that will be applied directly to the `td` or `th` for the column. Then we can add `cellProps={{ className: "w-16" }}` the column config (or an emotion variant for projects outside of engine-frontend) and that can be spread directly to the `th`s and `td`s. Then if we ever need other props to be added, the logic is already there.

What do you think @fmvfelipemonsalve ?

src/Table.tsx
```diff
@@ -82,26 +90,68 @@ export function Table<RowShape>({
...
+                        width:
+                          index % 5 === 0
+                            ? '66%'
+                            : index % 5 === 1 || index % 5 === 4
+                              ? '75%'
+                              : index % 5 === 2
+                                ? '60%'
+                                : '83%',
+                        borderRadius: '0.25rem',
+                        animationName: 'tw-keyframes-glow',
```
justinanastos: I am advocating for removing this loading logic, but I feel it's important to mention that this code lives in `space-kit` which has no knowledge of our tailwind classes, so `tw-keyframes-glow` will have no effect unless this is used in engine-frontend.

src/Table.tsx
justinanastos: Proposed update:

```ts
/**
 * Add props directly to this column's `td`
 *
 * Accepts either an object or a function taking the row data and returning
 * an object. The result will be spread directly to the `td` with the
 * `className` being merged in.
 */
tableDataProps?:
  | React.DetailedHTMLProps<
      React.HTMLAttributes<HTMLTableRowElement>,
      HTMLTableRowElement
    >
  | ((
      rowShape: RowShape
    ) => React.DetailedHTMLProps<
      React.HTMLAttributes<HTMLTableRowElement>,
      HTMLTableRowElement
    >);

/**
 * Add props to the header's `<th>`
 *
 * This is useful for configuring the size of the column
 */
tableHeaderProps?: React.DetailedHTMLProps<
  React.ThHTMLAttributes<HTMLTableHeaderCellElement>,
  HTMLTableHeaderCellElement
>;
```

justinanastos: Another option, perhaps even simpler, would be to use `<colgroup>` This also would need to be configured into the table, but would be done similarly:

https://www.w3schools.com/tags/tag_colgroup.asp

```tsx
/**
 * Pass these props to `col` elements nested below the `table`'s single
 * `<colgroup>`.
 *
 * This allows you to apply styles to columns (including setting sizes) by
 * setting a class on a single element instead of _all_ elements in a
 * table's row
 */
colProps?: React.DetailedHTMLProps<
  React.ColHTMLAttributes<HTMLTableColElement>,
  HTMLTableColElement
>;
```

src/Table.tsx
mayakoneval: I would guess loading bars are the preferred loading state, and that should definitely be possible to implement in engine rather than here as Justin mentioned. +1 for doing this in engine rather than space kit

src/Table.tsx
mayakoneval: After discussing with Justin, the colgroup option seems like a good way to go. Less code and you can pass span, or width etc as a consumer.

src/Table.tsx
mayakoneval: Since we already have all the animation logic in engine, lets keep it there!

fmvfelipemonsalve:
My guess is you were working on an old branch, since I saw the old file yesterday and removed it. Will force push again to fix the branch (with your commit on top), and you can pull. Sounds good?

src/Table/Table.tsx
```diff
@@ -60,6 +71,12 @@ export function Table<RowShape>({
         width: "100%",
       }}
     >
+      <colgroup>
+        {columns.map(({ colProps }, index) => (
+          <col key={index} {...colProps} />
```
justinanastos: Using an `index` for a `key` is an anti-pattern. I didn't even realize our linter allowed this. It won't matter here until we change the columns (like if we offer re-ordering of the columns, which we might do). There's always a better way. In this case, use the `column`'s `id`.

src/Table/Table.tsx
```diff
@@ -96,6 +113,8 @@ export function Table<RowShape>({
                       ? `none`
                       : `1px solid ${colors.silver.dark}`,
                   padding,
+                  paddingLeft: colIndex === 0 ? 0 : padding,
```
justinanastos: Why are you changing the padding here? I think @mayakoneval suggested you create a storybook story showing how to use the `colProps`, is that what you meant to do?

fmvfelipemonsalve: @justinanastos This was a commit @mayakoneval slipped into my PR haha. I assume she didn't want to create a whole new PR. But you're right in that I forgot to add the storybook to demo colProps. Will do that.

mayakoneval: hi I just wanted to make the the left was left aligned and the right could be right aligned so I'm taking out the padding because in all table designs this is the case!

src/Table/Table.story.tsx
```diff
@@ -101,3 +101,67 @@ storiesOf("Table", module)
       ]}
     />
   ));
+
+storiesOf("Table", module)
+  .addDecorator(withKnobs)
```
justinanastos: Unless you're using `knobs`, don't include this.

PR #33, opened 2019-07-29
Make Icon weight configurable

justinanastos:
First, this PR will look scary and that's ok.

When we generate the React files, we're given an AST of svgr's final output. We can then modify that to add or remove things. This will allow us to make the stroke width configurable and to set a style to prevent the size from causing the stroke width to change.

Take a look at the storybook story and see that it's now configurable with a weight!

### DIFF
--- a/src/icons/scripts/convert.ts
+++ b/src/icons/scripts/convert.ts
@@ -1,8 +1,10 @@
 import fs from "fs";
 import path from "path";
+import update from "immutability-helper";
 import svgr from "@svgr/core";
 import { formatComponentName } from "./formatComponentName";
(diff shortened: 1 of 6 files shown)


src/icons/scripts/convert.ts
```diff
@@ -42,6 +44,43 @@ const COMPONENT_PATH = path.resolve(__dirname, "..", "..", "..", "icons");
               path.basename(filename, path.extname(filename))
             );
 
+            function makeClassName() {
```
timbotnik: nit: it seems like this could just be a JS object instead of a function?  also it's making a _css_ property not so much a classname so perhaps `const svgCss = "...";  const cssProp = {...};`

PR #34, opened 2019-07-29
AP-703: Write integration tests (AP-780)

justinanastos:
This is a second pass at writing integration tests. We don't target any code coverage, we only test things that give us increased confidence. If the tests pass, we should be confident to release. Not all components are tested yet, we can add more tests as we go.

I also included some resources to read about testing and some very very rough guidelines on what I'd be looking for. Please open PRs to help tell our story!

Resolves [AP-703](https://apollographql.atlassian.net/browse/AP-703) and [AP-780](https://apollographql.atlassian.net/browse/AP-780)


src/Button.spec.tsx
```diff
@@ -0,0 +1,200 @@
...
+              theme={theme}
+              icon={<IconShip2 />}
+              data-testid="button"
+              size="small"
+            />
+          );
+          expect(getByTestId("button")).toBeInTheDocument();
+        });
+
+        it("with large size", () => {
```
mayakoneval: With these tests, we aren't checking the size of a large button, or the appearance of a flat button, for example. Is that something we can do with jest? Is that something we want to do here, or does chromatic cover that?

src/Table.spec.tsx
```diff
@@ -0,0 +1,232 @@
...
+    dateAdded: "05/28/2019",
+  },
+  {
+    name: "Jeremy Griffin",
+    email: "jgriffin@starkindustries.net",
+    image: require("./table.story/jeremy-griffin.png"),
+    dateAdded: "05/29/2019",
+  },
+];
+
```
mayakoneval: once this is merged, perhaps @fmvfelipemonsalve should consider adding tests for the width passing?

src/Button.spec.tsx
justinanastos: This is a great question. I need to document exactly what the expectations are for tests or no one will know what to write on their own.

These tests are intended to test _functionality_. That means when we pass stuff, those things are rendered as expected. All behaviors (like clicking the mouse or pressing Escape) should be tested.

These tests are incapable of testing visual appearance because they are never painted. Jest uses jsdom to render the dom; but it doesn't actually lay that out or paint it; it just produces a dom that we can analyze. We'll need to use Chromatic to visually test things.

One more note: these tests are expected to be fast. Visual testing is expected to be slow. I would expect that we can develop with TDD (write tests before writing code) and then watching jests's output as we work to make sure we haven't broken anything. I expect Chromatic to run in CI, not on local workstations because it's slow.

src/Table.spec.tsx
justinanastos: Yes :) When new functionality is added, tests should be added to prove that it works.

justinanastos:
> I'm really excited about this. I can't think of any cases that we have missed right now, but it should be standard for people opening space-kit PR's to add tests for each new prop / new component.
> 
> As long as we are of the mind that these ci / jest tests cover components being rendered and chromatic covers the appearance of components, I have confidence that a passing PR can be merged without issues.

Something else I want to check: are any tests superfluous? 

Also, should we stop nesting like KCD wrote about today (https://kentcdodds.com/blog/avoid-nesting-when-youre-testing/)? I think yes.

apollo-bot2:
:rocket: PR was released in v2.3.0 :rocket:

PR #35, opened 2019-07-30
Export `colors` as a constant

justinanastos:
This will require we change how we import `colors`.

Old:

```jsx
import * as colors from '@apollo/space-kit/colors';
```

New:

```jsx
import { colors } from '@apollo/space-kit/colors';
```

The `import *` is awkward and it's a pain to remember when we need that. So let's never do it.


timbotnik:
i think this negates any possibility of tree-shaking based on only importing a _subset_ of colors, but everyone always doing `import *` coupled with the fact that we are only talking about a few constant definitions indicates to me that this is not a very serious objection.

PR #40, opened 2019-07-31
Card component

mayakoneval:
- Card component
- Section component image 
  - large / standard size
 - Card story

**Props**

- **children?: React.ReactNode;**
   card content, aligned below title and description and title chlildren

- **title?: React.ReactNode;**
title of the card
- **className?: string;**

- **description?: React.ReactNode;**
same size grey text under the title as seen in designs

- **titleChildren?: React.ReactNode;**
exists for things like left centered buttons and text, as seen in the designs below

- **forceNoChildPadding?: boolean;**
exists only for cards with hidden children (like drawers) to force no bottom padding

- **size?: "standard" | "large";**
two size cards, one for things like the [user management table,](https://app.zeplin.io/project/5b3665ece41d8c86a0f30286/screen/5d095654b2c0555e11b4cba7) and one for [standard cards ](https://app.zeplin.io/project/5b3665ece41d8c86a0f30286/screen/5d3a355f62259b6cb6021358) difference is the size of the title text and the padding

[designs (one of many)](https://app.zeplin.io/project/5b3665ece41d8c86a0f30286/screen/5d0957502ac55d5d2c988ec7)

**TODO:**
- add tests when[ 34](https://github.com/apollographql/space-kit/pull/34) is merged

Questions:
- How organized do we want our storybook? I just added tabs for the cards but they aren't the same sort of feel as the designed and structured stories for buttons, and I never implemented the spec'd storybook for modals. This was because I prioritized getting to use the components in engine.
- right now I have a card section which looks like this 
image and it adds padding to the top whenever you use it, however, if you use it right after a card it will add double padding (one from the card, one from the section) so I have to specify `forceNoChildPadding`. There is definitely a more elegant way to do this, but I will keep thinking about it

### DIFF
--- a/src/Card/Card.story.tsx
+++ b/src/Card/Card.story.tsx
@@ -2,7 +2,7 @@
 import { jsx } from "@emotion/core";
 import { useState } from "react";
 import { storiesOf } from "@storybook/react";
-import { Card } from "./Card";
+import { Card, CardSection } from "./Card";
 import { Button } from "../Button";
 import * as colors from "../colors";
 
@@ -104,5 +104,40 @@ storiesOf("Card", module)
       }}
     >
       <CardWithDrawer />
+    </div>
+  ))
+  .add("multiple sections", () => (
+    <div
+      css={{
+        backgroundColor: colors.silver.light,
+        padding: "40px",
+        height: "100vh",
+      }}
+    >
+      <Card
+        title="Card Title"
+        description="description goes here"
+        titleChildren={
+          <Button color={colors.red.base}>
+            <div css={{ color: colors.white }}>Click</div>
+          </Button>
+        }
+        forceNoChildPadding={true}
+      >
+        <CardSection
+          title="Members"
+          description="You have 7 members in your organization"
+        />
+        <CardSection
+          title="Members"
+          description="You have 7 members in your organization"
+          border={true}
+          titleChildren={
+            <Button color={colors.red.base}>
+              <div css={{ color: colors.white }}>Edit</div>
+            </Button>
+          }
+        />
+      </Card>
     </div>
   ));
--- a/src/Card/Card.story.tsx
+++ b/src/Card/Card.story.tsx
@@ -1,0 +1,108 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core";
+import { useState } from "react";
+import { storiesOf } from "@storybook/react";
+import { Card } from "./Card";
+import { Button } from "../Button";
+import * as colors from "../colors";
+
+const CardWithDrawer = () => {
+  const [drawerOpen, setDrawerOpen] = useState<boolean>(false);
+  return (
+    <Card
+      title="Card Title"
+      description="description goes here"
+      forceNoChildPadding={true}
+      titleChildren={
+        <Button
+          color={colors.red.base}
+          onClick={() => setDrawerOpen(!drawerOpen)}
+        >
+          <div css={{ color: colors.white }}>Click</div>
+        </Button>
+      }
+    >
+      {drawerOpen && (
+        <div
+          css={{
+            height: "2.5rem",
+            backgroundColor: colors.black.base,
+            color: colors.white,
+            padding: 20,
+          }}
+        >
+          card content
+        </div>
+      )}
+    </Card>
+  );
+};
+
+storiesOf("Card", module)
+  .add("default", () => (
+    <div
+      css={{
+        backgroundColor: colors.silver.light,
+        padding: "40px",
+        height: "100vh",
+      }}
+    >
+      <Card title="Card Title" description="description goes here">
+        <div>card content</div>
+      </Card>
+    </div>
+  ))
+  .add("large", () => (
+    <div
+      css={{
+        backgroundColor: colors.silver.light,
+        padding: "40px",
+        height: "100vh",
+      }}
+    >
+      <Card
+        size="large"
+        title="Card Title"
+        description="description goes here"
+        titleChildren={
+          <Button color={colors.red.base}>
+            <div css={{ color: colors.white }}>Click</div>
+          </Button>
+        }
+      >
+        <div>card content</div>
+      </Card>
+    </div>
+  ))
+  .add("with title children - button", () => (
+    <div
+      css={{
+        backgroundColor: colors.silver.light,
+        padding: "40px",
+        height: "100vh",
+      }}
+    >
+      <Card
+        title="Card Title"
+        description="description goes here"
+        titleChildren={
+          <Button color={colors.red.base}>
+            <div css={{ color: colors.white }}>Click</div>
+          </Button>
+        }
+      >
+        <div>card content</div>
+      </Card>
+    </div>
+  ))
+  .add("force no child padding (use for drawers)", () => (
+    <div
+      css={{
+        backgroundColor: colors.silver.light,
+        padding: "40px",
+        height: "100vh",
+      }}
+    >
+      <CardWithDrawer />
+    </div>
+  ));
--- a/src/Card/Card.story.tsx
+++ b/src/Card/Card.story.tsx
@@ -3,10 +3,21 @@ import { jsx } from "@emotion/core";
 import { storiesOf } from "@storybook/react";
 import { Card, CardSection, CardSeperator } from "./Card";
 import { Button } from "../Button";
-import * as colors from "../colors";
+import { colors } from "../colors";
 
 storiesOf("Card", module)
   .add("default", () => (
+    <div
+      css={{
+        backgroundColor: colors.silver.light,
+        padding: "40px",
+        height: "100vh",
+      }}
+    >
+      <Card />
+    </div>
+  ))
+  .add("default, heading, description", () => (
     <div
       css={{
         backgroundColor: colors.silver.light,
--- a/src/Card/Card.story.tsx
+++ b/src/Card/Card.story.tsx
@@ -1,42 +1,9 @@
 /** @jsx jsx */
 import { jsx } from "@emotion/core";
-import { useState } from "react";
 import { storiesOf } from "@storybook/react";
-import { Card, CardSection } from "./Card";
+import { Card, CardSection, CardSeperator } from "./Card";
 import { Button } from "../Button";
 import * as colors from "../colors";
-
-const CardWithDrawer = () => {
-  const [drawerOpen, setDrawerOpen] = useState<boolean>(false);
-  return (
-    <Card
-      title="Card Title"
-      description="description goes here"
-      forceNoChildPadding={true}
-      titleChildren={
-        <Button
-          color={colors.red.base}
-          onClick={() => setDrawerOpen(!drawerOpen)}
-        >
-          <div css={{ color: colors.white }}>Click</div>
-        </Button>
-      }
-    >
-      {drawerOpen && (
-        <div
-          css={{
-            height: "2.5rem",
-            backgroundColor: colors.black.base,
-            color: colors.white,
-            padding: 20,
-          }}
-        >
-          card content
-        </div>
-      )}
-    </Card>
-  );
-};
 
 storiesOf("Card", module)
   .add("default", () => (
@@ -47,7 +14,7 @@ storiesOf("Card", module)
         height: "100vh",
       }}
     >
-      <Card title="Card Title" description="description goes here">
+      <Card heading="Card heading" description="description goes here">
         <div>card content</div>
       </Card>
     </div>
@@ -62,9 +29,9 @@ storiesOf("Card", module)
     >
       <Card
         size="large"
-        title="Card Title"
+        heading="Card heading"
         description="description goes here"
-        titleChildren={
+        actions={
           <Button color={colors.red.base}>
             <div css={{ color: colors.white }}>Click</div>
           </Button>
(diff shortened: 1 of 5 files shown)


cheapsteak:
Will this component (specifically `titleChildren`) be able to accommodate this card?
image   

I might have recommended a more compound-component-y approach (e.g. [semantic-ui-react's Card](https://react.semantic-ui.com/views/card/#types-card) (click on "try it" on one of the examples)) to try to break the component down into compose-able pieces rather than adding props to account for variations (to avoid needing props like `forceNoChildPadding` and `titleChildren`), but will defer to the others on if we would prefer to ship-fast-and-iterate    

---

Side-note, for the props list in the PR description, I'd recommend adding a bit of separation between each prop (bullet point could work), and move the description of a prop after the prop to improve readability, e.g.

  - **children?: React.ReactNode;**
    card content, aligned below title and description and title chlildren
  - **title?: React.ReactNode;**
    title of the card
  - **className?: string;**
  - **description?: React.ReactNode;**
    same size grey text under the title as seen in designs
  - **titleChildren?: React.ReactNode;**
    exists for things like left centered buttons and text, as seen in the designs below
  - **forceNoChildPadding?: boolean;**
    exists only for cards with hidden children (like drawers) to force no bottom padding
  - **size?: "standard" | "large";**
    two size cards, one for things like the [user management table,](https://app.zeplin.io/project/5b3665ece41d8c86a0f30286/screen/5d095654b2c0555e11b4cba7) and one for [standard cards ](https://app.zeplin.io/project/5b3665ece41d8c86a0f30286/screen/5d3a355f62259b6cb6021358) difference is the size of the title text and the padding

mayakoneval:
> Will this component (specifically `titleChildren`) be able to accommodate this design?
> image
> 
> I might have recommended a more compound-component-y approach (e.g. [semantic-ui-react's Card](https://react.semantic-ui.com/views/card/#types-card) (click "try it" on one of the examples)) to try to break the component down into compose-able pieces rather than adding props to account for variations (to avoid needing props like `forceNoChildPadding` and `titleChildren`), and err towards abstracting less, but will defer to the others on if we would prefer to ship-fast-and-iterate
> 
> Side-note, for the props list in the PR description, I'd recommend adding a bit of separation between each prop (bullet point could work), and move the description of a prop after the prop to improve readability, e.g.
> 
> * **children?: React.ReactNode;**
>   card content, aligned below title and description and title chlildren
> * **title?: React.ReactNode;**
>   title of the card
> * **className?: string;**
> * **description?: React.ReactNode;**
>   same size grey text under the title as seen in designs
> * **titleChildren?: React.ReactNode;**
>   exists for things like left centered buttons and text, as seen in the designs below
> * **forceNoChildPadding?: boolean;**
>   exists only for cards with hidden children (like drawers) to force no bottom padding
> * **size?: "standard" | "large";**
>   two size cards, one for things like the [user management table,](https://app.zeplin.io/project/5b3665ece41d8c86a0f30286/screen/5d095654b2c0555e11b4cba7) and one for [standard cards ](https://app.zeplin.io/project/5b3665ece41d8c86a0f30286/screen/5d3a355f62259b6cb6021358) difference is the size of the title text and the padding
> 
> (appreciate you adding this though, helped me understand what titleChildren was for)

I like this idea, I was trying to do an 80% implementation for most of the org settings stuff. This would get rid of a lot of repeated code where we have black titles and grey descriptions, so I think at the least there should be a card.section which has a border and the title / description / titleChildren props.

src/Card/Card.tsx
```diff
@@ -0,0 +1,100 @@
...
+  size,
+  ...otherProps
+}) => (
+  <div
+    {...otherProps}
+    css={{
+      backgroundColor: colors.white,
+      color: colors.black.base,
+      boxShadow: `0 4px 8px 0 rgba(0, 0, 0, .04)`,
+      borderRadius: `0.5rem`,
```
justinanastos: I don't think should be in `rem`s because border radii probably shouldn't change if the page's font size changes.

src/Card/Card.tsx
```diff
@@ -0,0 +1,100 @@
...
+      paddingLeft: 24,
+      paddingRight: 24,
+      margin: "2rem",
+      paddingTop: size === "large" ? 16 : 28,
+      paddingBottom: size === "large" ? 16 : 28,
+    }}
+  >
+    <div css={{ display: "flex" }}>
+      <div css={{ flex: "1 1 0%" }}>
+        <div css={{ display: "block" }}>
```
justinanastos: Why is this `display: "block"`, isn't that the default for a `div`?

src/Card/Card.tsx
```diff
@@ -0,0 +1,100 @@
...
+          {title && (
+            <div
+              css={{
+                display: "flex",
+                color: colors.black.base,
+                ...(size === "large" ? base.xxlarge : base.large),
```
justinanastos: Awesome use of `...()` :-D

src/Card/Card.tsx
```diff
@@ -0,0 +1,100 @@
...
+              }}
+            >
+              <span
+                css={{
+                  fontWeight: 600,
```
justinanastos: Does this size conform to something in our typography? If not, then we should talk to @adamatorres about changing it.

src/Card/Card.tsx
```diff
@@ -0,0 +1,100 @@
...
+                  flex: "1 1 0%",
+                  lineHeight: 1.5,
+                  overflow: "hidden",
+                  textOverflow: "ellipsis",
+                  whiteSpace: "nowrap",
+                  paddingRight: "1rem",
```
justinanastos: I think we need to avoid `rem`s now until we figure out exactly how we're going to make adjustments with the page's font-size.

src/Card/Card.tsx
```diff
@@ -0,0 +1,100 @@
...
+                  !!children && !forceNoChildPadding ? "1.5rem" : "",
+              }}
+            >
+              {description}
+            </div>
+          )}
+        </div>
+      </div>
+      {titleChildren && (
+        <div css={{ flex: "none", marginLeft: "1rem" }}>{titleChildren}</div>
```
justinanastos: Do we want to automatically align this to the right? We can do that by making the _other_ child in the flexbox have a `margin-right` of `auto`.

src/Card/Card.tsx
```diff
@@ -0,0 +1,100 @@
+/** @jsx jsx */
+import * as colors from "../colors";
+import { base } from "../typography";
+import { jsx } from "@emotion/core";
+import React from "react";
+import PropTypes from "prop-types";
+
+interface Props {
```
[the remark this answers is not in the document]
mayakoneval: @justinanastos is this just a request to delete the documentation for trivial things like the title, or also to add more documentation for child? Just want to clarify.

src/Card/Card.tsx
```diff
@@ -0,0 +1,100 @@
...
+  children?: React.ReactNode;
+  title?: React.ReactNode;
+  className?: string;
+  description?: React.ReactNode;
+  titleChildren?: React.ReactNode;
+  forceNoChildPadding?: boolean;
```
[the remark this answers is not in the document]
mayakoneval: This exists because, if there are children, we want to add padding between the title and the children. Sometimes, cards have drawers which are passed as children, but hidden until they are triggered. I'm fine including the double padding for drawers since we are getting rid of drawers in our future engine designs.

src/Card/Card.tsx
mayakoneval: this corresponds to the base typography, which doesn't have a set fontWeight so we override it to make it semibold. The same thing is done [here in button](https://github.com/apollographql/space-kit/pull/26/files#diff-9b4a3ce6a1b69789a02162483b7f8141R146)

justinanastos:
This PR reminds me again of why we shouldn't manually create documentation. You changed the props in the PR but not in the description :)

src/Card/Card.tsx
justinanastos: Why are you also overriding the `lineHeight`?

src/Card/Card.tsx
```diff
@@ -0,0 +1,192 @@
...
+                {heading}
+              </span>
+            </div>
+          )}
+          {description && (
+            <div
+              css={{
+                ...base.base,
+                color: colors.grey.base,
+                maxWidth: actions ? 650 : "",
```
justinanastos: Where did this magic number (650) come from?

src/Card/Card.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+Card.propTypes = {
+  children: PropTypes.node,
+  heading: PropTypes.node,
+  description: PropTypes.node,
+  actions: PropTypes.node,
+  size: PropTypes.oneOf<Props["size"]>(["standard", "large"]),
+};
+
+export const CardSection: React.FunctionComponent<Props> = ({
```
justinanastos: NBD: use the `React.FC` shortcut

(NBD = your call, only change it if you want to)

src/Card/Card.tsx
```diff
@@ -0,0 +1,192 @@
...
+  actions,
+}) => (
+  <div
+    css={{
+      display: "flex",
+      marginTop: 24,
+    }}
+  >
+    <div css={{ flex: "1 1 0%", marginRight: "auto" }}>
+      <div css={{ display: "block" }}>
```
justinanastos: Why does this need a `display: block` if that's the default for a `div`?

src/Card/Card.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+CardSection.propTypes = {
+  heading: PropTypes.node,
+  description: PropTypes.node,
+  actions: PropTypes.node,
+};
+
+export const CardSeperator: React.FunctionComponent = () => (
+  <hr
+    style={{
```
justinanastos: Use `css` instead of `style`.

src/Card/Card.tsx
```diff
@@ -0,0 +1,192 @@
...
+  </div>
+);
+
+CardSection.propTypes = {
+  heading: PropTypes.node,
+  description: PropTypes.node,
+  actions: PropTypes.node,
+};
+
+export const CardSeperator: React.FunctionComponent = () => (
```
justinanastos: NBD: `React.FC`

src/Card/Card.tsx
justinanastos: Add a `/** */` explanation of what a `CardSection` is. We can use this in the autogenerated docs later.

src/Card/Card.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * actions could be a button
+   * or a tooltip or anything the card should display
+   * aligned with the title on the right
+   */
+  actions?: React.ReactNode;
+
+  /**
+   * cards can be standard (18px text, larger padding)
```
justinanastos: NIT: I'd leave out the 18px text here because then you have to remember to change it in two places. If you change it below and not here, you have outdated documentation, which is worse than no documentation. `size` can be self documenting too :) 

Push back when you think things don't make sense!

src/Card/Card.story.tsx
```diff
@@ -0,0 +1,98 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core";
+import { storiesOf } from "@storybook/react";
+import { Card, CardSection, CardSeperator } from "./Card";
+import { Button } from "../Button";
+import * as colors from "../colors";
+
+storiesOf("Card", module)
+  .add("default", () => (
```
justinanastos: Create a story with no description and no title

Think of stories as having two purposes:

1. To develop (as you know)
2. To validate for visual regressions in chromatic (which we haven't added yet, but we will!)

src/Card/Card.tsx
mayakoneval: its the equivalent of max-w-2xl in tailwind I believe, I will make const

src/Card/Card.tsx
```diff
@@ -0,0 +1,191 @@
+/** @jsx jsx */
+import { colors } from "../colors";
+import { base } from "../typography";
+import { jsx } from "@emotion/core";
+import React from "react";
+import PropTypes from "prop-types";
+
+const descriptionMaxWidth = 650;
```
timbotnik: perhaps `640` if we are on a 4px grid 😁

src/Card/Card.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+Card.propTypes = {
+  children: PropTypes.node,
+  heading: PropTypes.node,
+  description: PropTypes.node,
+  actions: PropTypes.node,
+  size: PropTypes.oneOf<Props["size"]>(["standard", "large"]),
+};
+
+export const CardSection: React.FC<Props> = ({
```
timbotnik: It looks like a `CardSection` has less props than a `Card` - we should probably split these into separate `CardProps` and `CardSectionProps`.

src/Card/Card.tsx
```diff
@@ -0,0 +1,191 @@
...
+  description?: React.ReactNode;
+
+  /**
+   * actions could be a button
+   * or a tooltip or anything the card should display
+   * aligned with the title on the right
+   */
+  actions?: React.ReactNode;
+
+  size?: "standard" | "large";
```
timbotnik: Maybe a comment here about how this increases the padding and heading font size would be nice.

src/Card/Card.tsx
timbotnik: As per [Slack convo](https://meteor.slack.com/archives/CJM405HJB/p1565054671017800) - we're going to export additional typographies that include the font weight option.  IMO we can do that in a separate PR.

PR #45, opened 2019-08-03
AP-544 Implement Text Inputs

mayakoneval:
This PR includes: 
- Text Inputs 
- Steppers
- Text Inputs with icons

[Input Props
](https://github.com/apollographql/space-kit/pull/45/files#diff-198962a3c387a9ddec485eb291f6497dR8)

[Stepper Props
](https://github.com/apollographql/space-kit/pull/45/files#diff-89cd8aee1ba9a409566aabfdfdb71012R16)

[Design](https://app.zeplin.io/project/5c7dcb5ab4e654bca8cde54d/screen/5cfeb2b92d682519cad41830)

[Ticket](https://apollographql.atlassian.net/browse/AP-544)

[Design Quip Doc](https://apollographql.quip.com/Cnh7AXu7Sp5d/Space-Kit-Text-Input-Design-System)

Points of conversation:
- The `Stepper` `onClick` accepts a number, while the `Input` `onClick` accepts an `HTMLInputEvent`. This is because the Stepper should only call `onChange` when the number has actually changed, not when the user makes an action in the input. For example, they could try to decrease the number but the number is at min already, so we don't want to call `onClick`.
- The `Stepper` is copy pasted from Tim's implementation and then emotionified. However, I took the responsibility of updating the value away from the consumer and into the component. @timbotnik curious as to what you think of this. 
- The PR **does not** include input masking, as Recurly does that for us when we need it. 
- @adamatorres the error and info states look like this (not filled in icons)

image

Is that okay? Note the icons are bottom aligned with the text, do you want the icons aligned `display:flex` style with the text so that when it wraps the icon is centered to the text like this? 

image 

Or would you rather it be in line with the text?


src/Input/InputWrapper.tsx
```diff
@@ -0,0 +1,121 @@
...
+              marginTop: 8,
+              marginRight: 8,
+              paddingLeft: size === "small" ? 8 : 12,
+              color: error ? colors.red.base : colors.grey.base,
+              ...typography.base.small,
+            }}
+          >
+            {error ? (
+              <span>
+                <IconError css={{ marginRight: 8, width: 14, height: 14 }} />
```
cheapsteak: To match design, it looks like the icon should be a separate column from the text and aligned to the center of the top line of text

src/Input/Stepper.tsx
```diff
@@ -0,0 +1,257 @@
...
+              css={{
+                width: 10,
+                height: 10,
+                color:
+                  disabled || atMin ? colors.grey.lighter : colors.blue.base,
+              }}
+            />
+          </button>
+
+          <input
```
cheapsteak: let's add a `type="number"` attribute

src/Input/Input.tsx
```diff
@@ -0,0 +1,128 @@
...
+          ":hover,  &[data-force-hover-state]": {
+            borderColor:
+              !disabled && !error
+                ? colors.grey.light
+                : error
+                ? colors.red.base
+                : colors.silver.darker,
+          },
+        }}
+        type="text"
```
cheapsteak: 'text' is the default type and doesn't need to be passed  
having this here after `{...otherProps}` means the type can't be overridden to number/date/email

src/Input/InputWrapper.tsx
```diff
@@ -0,0 +1,121 @@
...
+  description,
+  size = "standard",
+  error,
+  info,
+  icon,
+  children,
+}: Props): ReturnType<React.FC> {
+  return (
+    <React.Fragment>
+      <label
```
cheapsteak: We might want to have `<label/>` tag wrap the `{label}`, `{description}`, and `{children}` (not the error or info messages though), it would automatically associate the text we're using to describe the input with the input box for accessibility ([more info](https://stackoverflow.com/questions/774054/should-i-put-input-elements-inside-a-label-element)). It'd also give focus to the input when clicking on the title, which is imo desirable  

Alternatively, we could use `aria-label` or `aria-labelledby`, but those are more of a hassle (need to associate them by id, generating unique ids for association is a bit more of a hassle than associating via nesting hierarchy)

src/Input/InputWrapper.tsx
```diff
@@ -0,0 +1,121 @@
...
+              ...typography.base.small,
+            }}
+          >
+            {error ? (
+              <span>
+                <IconError css={{ marginRight: 8, width: 14, height: 14 }} />
+              </span>
+            ) : (
+              <span>
+                <IconInfo css={{ marginRight: 8, width: 14, height: 14 }} />
```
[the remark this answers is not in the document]
timbotnik: Agree this API is a little bit restrictive.  RE: solutions, I'm happy to ship with this original version here and wait to see what other things we need.  I think limiting the value to a text-only here might be an issue, I'm especially thinking about links to documentation etc, which would point me in direction of the `helper` prop idea.  I think I would prefer to call it a `message` prop?  `<Input message={<ErrorMessage>message</ErrorMessage>}>` which leaves us able to create a warning message etc.

src/Input/Stepper.tsx
```diff
@@ -0,0 +1,257 @@
...
+  disabled,
+  description,
+  label,
+  size = "standard",
+  error,
+  info,
+  ...otherProps
+}) => {
+  const [focused, setFocused] = useState<boolean>(false);
+  const [value, setValue] = useState<number>(initialValue || 10);
```
timbotnik: hmm, 10 is a funny choice 😁 let's just pass thru `initialValue`.

src/Input/Stepper.tsx
```diff
@@ -0,0 +1,257 @@
...
+    if (newValue !== undefined && newValue < min) {
+      setValue(min);
+      if (onLimitReached) onLimitReached("min");
+    } else if (newValue !== undefined && newValue > max) {
+      setValue(max);
+      if (onLimitReached) onLimitReached("max");
+    } else if (newValue !== value) {
+      setValue(newValue);
+    }
+    onChange(value);
```
timbotnik: I think this should be `newValue`?

src/Input/Stepper.tsx
```diff
@@ -0,0 +1,257 @@
...
+    setFocused(false);
+    handleChange(value);
+  };
+
+  const onInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const newValue = parseInt(event.target.value, 10);
+    if (!isNaN(newValue)) {
+      handleChange(newValue);
+    } else {
+      handleChange(0);
```
timbotnik: I don't think it makes sense to fire this with a 0 if the value is not a number.  `undefined` makes more sense to me.

src/Input/Stepper.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+          <button
+            type="button"
+            onClick={onIncrement}
+            disabled={atMax || disabled}
+            css={{
+              paddingLeft: 4,
+              paddingRight: 4,
+              margin: 4,
+              fontWeight: 300,
```
timbotnik: setting `fontWeight` here around the SVG is a no-op yeah?

src/Input/Stepper.tsx
```diff
@@ -0,0 +1,257 @@
...
+      </div>
+    </InputWrapper>
+  );
+};
+Stepper.propTypes = {
+  initialValue: PropTypes.number,
+  onChange: PropTypes.func.isRequired,
+  onLimitReached: PropTypes.func,
+  min: PropTypes.number,
+  max: PropTypes.number,
```
timbotnik: Missing some propTypes from `Props`?

src/Input/Stepper.tsx
mayakoneval: We only want to call onChange with the `value` if it changes, and in this case `newValue` is the value the user has just tried to make the stepper. `value` willonly change if thats a valid value, so I call `onChange(value)` after all the logic determining if `newValue` is valid is done and value is set. Does that make sense?

src/Input/Stepper.tsx
mayakoneval: whoops!

src/Input/InputWrapper.tsx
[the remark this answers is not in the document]
mayakoneval: @cheapsteak @timbotnik agreed, I will export `ErrorMessage` and `InputMessage` and allow the helperText to be a react node. Question for you: do you think it makes sense for the `ErrorMessage` and `InputMessage` to live in the same file as `Input`, if so should `Stepper` be moved into `Input`?

I am imagining a scenario where someone wants a stepper with info messages and has to import the messaging from Input.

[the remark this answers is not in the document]
mayakoneval: I am realizing that we need at least the error because the border color of the input is set this way, I will add an error boolean.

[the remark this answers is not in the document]
cheapsteak: One thing we _could_ do is to have the code for those components be in separate files, and re export them from both Input and Stepper  

What think?

adamatorres:
Hey @mayakoneval -- Here's some design feedback on the inputs:
- [ ] text size in the small input should be 13px, looks like it's 15px. 
- [ ] placeholder text color `grey-light`, it looks a bit darker than it should
- [ ] disabled place holder text is too dark, it should be `grey-lighter`
- [ ] Helper text needs to support an icon (specifically in error cases) and no icon. We currently don't have an example of basic helper text without an icon.
- [ ] When helper text below an input has an icon and the text wraps, we shouldn't wrap the text under the icon.
- [ ] Let's use the filled icons with the helper text.
- [ ] I'm also seeing a delay of the focus state of the input after clicking into it, not sure what's going on there, but it should feel more instantaneous.

jgzuke:
A few thoughts from using the new Inputs.
- Its really awesome to have error states built in 🎉It'll make everything a lot more consistent and easier to build, I got to kill a bunch of code
- Its a bit weird to have the input manage its own state instead of being a true managed/controlled component. It makes certain things impossible to use the new inputs with Ie recurly inputs (because recurly does dom manipulations), but it also means if the parent keeps the value in its own state as well we have extra render triggers.
- Right now we pass the `otherProps` (`className`s etc) to the inner input component which makes sense, however it means we cant pass any structure/position changing class names (`w-full` etc) since the input is wrapped. We can still position them by wrapping in divs or something, but it might make sense to take both props for the input, and for the outer wrapper

mayakoneval:
> A few thoughts from using the new Inputs.
> 
> * Its really awesome to have error states built in 🎉It'll make everything a lot more consistent and easier to build, I got to kill a bunch of code
> * Its a bit weird to have the input manage its own state instead of being a true managed/controlled component. It makes certain things impossible to use the new inputs with Ie recurly inputs (because recurly does dom manipulations), but it also means if the parent keeps the value in its own state as well we have extra render triggers.
> * Right now we pass the `otherProps` (`className`s etc) to the inner input component which makes sense, however it means we cant pass any structure/position changing class names (`w-full` etc) since the input is wrapped. We can still position them by wrapping in divs or something, but it might make sense to take both props for the input, and for the outer wrapper. Or we could have the `input` live as a sibling to the `label` instead of inside, which might fix this problem

I believe I fixed this with `inputProps` on the input and spreading the rest of the props on the outer div. Let me know if I'm off target :)

src/Input/InputWrapper.tsx
```diff
@@ -0,0 +1,132 @@
...
+  icon,
+  error,
+  children,
+  ...otherProps
+}: Props): ReturnType<React.FC> {
+  return (
+    <div {...otherProps}>
+      <label
+        css={{
+          paddingBottom: 8,
```
jgzuke: Imo in a component library, a components box should be no larger than it needs to be, padding/margins should be set by the people using the components. However, this is a lot less annoying once https://github.com/mdg-private/engine-frontend/pull/1577 or something analogous lands so 🤷‍♂

mayakoneval: @jgzuke From our Space Kit meetings, it seems that all apollo products want the same spacing / font etc and that is why we decided to create a component library, to be able to use components that already have padding / margins etc and have builtin descriptions / error states.

jgzuke: @mayakoneval the designs in https://app.zeplin.io/project/5d375cd742170577784b28ca/screen/5d41d60764e8c22bd5aeac20 have a smaller padding between them than two stacked inputs (with no labels)

justinanastos:
:rocket: PR was released in v1.4.0 :rocket:

package.json
```diff
@@ -13,6 +13,7 @@
     "clean:button": "rimraf Button",
     "clean:colors": "rimraf colors",
     "clean:fonts": "rimraf fonts",
+    "clean:formField": "rimraf FormField",
```
mayakoneval: There is no FormField anymore, so this npm script won't do anything.

CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ _From #57_
 - Edit Modal prop descriptions, fix margin top on children [#41](https://github.com/apollographql/space-kit/pull/41)
 - Export constant from `colors` instead of requiring `import * as colors` (#35)
 - Add Card component (#34)
+- Text Inputs and Steppers [(#45)](https://github.com/apollographql/space-kit/pull/45)
```
mayakoneval: This doesn't include Steppers, that file has been deleted

src/TextField/TextField.tsx
```diff
@@ -0,0 +1,232 @@
...
+   * Icon to have at the left of the input
+   */
+  icon?: React.ReactNode;
+
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+
+  /**
+   * Visible title
+   */
+  label: React.ReactNode;
```
mayakoneval: I don't think this should be necessary. Credit card form, for example

PR #46, opened 2019-08-05
AP-700 Export `emotionCacheProviderFactory` so allow consumers to override classes

justinanastos:
Ideally we'd want to use `<CacheProvder>` in the consumer application just like we'd want to use `<ApolloProvider>`, but a bug in emotion prevents this: https://github.com/emotion-js/emotion/issues/1386

This provides a `emotionCacheProviderFactory` factory that consumers can use to create a `CacheProvider`.

Contributes to [AP-700](https://golinks.io/j/AP-700)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ tsconfig.tsbuildinfo
 # Other modules
 /Button
 /colors
+/emotionCacheProviderFactory
 /fonts
 /icons
 /Modal
--- a/README.md
+++ b/README.md
@@ -59,26 +59,33 @@ function MyComponent() {
 
 ### Emotion
 
-Some components are styled with [emotion](https://emotion.sh) under the hood; emotion appends `<style>` tags to your `<head>` element at runtime. This may cause emotions's styles to be included _after_ your project's styles, preventing you from using your own classes to override emotion's. To get around this, you need to modify how emotion adds styles to the DOM. Check out the [docs](https://emotion.sh/docs/@emotion/cache). Here's an example:
+Some components are styled with [emotion](https://emotion.sh) under the hood; emotion appends `<style>` tags to your `<head>` element at runtime. This may cause emotions's styles to be included _after_ your project's styules, preventing you from using your own classes to override emotion's. 
 
-```ts
-import React from "react";
-import { CacheProvider } from "@emotion/core";
-import createCache from "@emotion/cache";
-import { RealApp } from "./RealApp";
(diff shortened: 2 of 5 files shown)


trevorblades:
In the issue that you linked, they're talking about the issue potentially being caused by multiple instances of emotion being loaded. I noticed that this library has @emotion/core as a dependency. Maybe it should be listed as a peer dependency instead, so that the consumer is responsible for installing only one version of emotion.

I remember seeing similar issues when a library was published with react as a dependency rather than a peer dependency. What do you think? Could that be the case here?

timbotnik:
> In the issue that you linked, they're talking about the issue potentially being caused by multiple instances of emotion being loaded. I noticed that this library has @emotion/core as a dependency. Maybe it should be listed as a peer dependency instead, so that the consumer is responsible for installing only one version of emotion.
> 
> I remember seeing similar issues when a library was published with react as a dependency rather than a peer dependency. What do you think? Could that be the case here?

I was thinking something similar, but happy to also go with this solution.

cheapsteak:
Hope that works  
If it doesn't, would it make sense to try exporting a `<SpaceKitProvider>` that for the moment is just a Cache Provider?

justinanastos:
I tried using a peer dependency and found that the issue persisted. If someone else wants to take a look and prove me wrong, I'd very much appreciate it.

cheapsteak:
Exporting the provider from space kit and importing it in engine front end "worked" for me

I was a bit surprised that it was injecting style tags into the container (style tags into the style tag, rather than style rules into the style tag). It does seem to be working as advertised (my mistake with the assumption). I think we can work with that. Will fiddle with it a bit in the morning

cheapsteak:
Exporting the provider and importing it in Engine Frontend _does_ work  

I debugged by replacing the entirety of AppLayout's return value in engine-frontend with

image

(`<SpaceKitProvider>` is just a renamed `<EmotionCacheProvider>` in this PR, same contents)

But, styles are placed nested inside the style tag 

image

Which feels weird. Everything works though  

image
image

Will defer to you folks on which way feels less weird

justinanastos:
I hear you on that being weird @cheapsteak 	; I only went with that because that's how their docs suggest doing it. I'd much rather have a `CacheProvider` that's handled externally but I haven't figured out how to do that yet 🤷‍♂

cheapsteak:
@justinanastos 	I was referring to the other solution I found still ending up weird :) 
 
Since it specifies a "container" rather than a position, it'd injects style tags into whatever the container is, but there aren't really any valid containers that can both hold `<style>` elements and live in `<head>`. It still works but works kinda weird

justinanastos:
> Since it specifies a "container" rather than a position, it'd injects style tags into whatever the container is, but there aren't really any valid containers that can both hold `<style>` elements and live in `<head>`. It still works but works kinda weird

You're totally right @cheapsteak; it's def weird.

justinanastos:
:rocket: PR was released in v1.1.0 :rocket:

PR #57, opened 2019-08-08
Refactor icon generation

justinanastos:
## Release Notes

* Refactor icon conversion script to use `@babel/types` instead of `immutability-helper` and trying to craft AST by hand. That was too opaque and it wasn't type safe. Now we're safe and it's as concise as it's going to get.

* Move intermediate icon files to `src/icons`

    This makes them accessible from anywhere else _and_ gives them access to other modules. We _were_ going to use this for the `EmotionProvider`, but now we aren't. This is less elegant and clever, but makes more sense and will make our code actually work.


justinanastos:
:rocket: PR was released in v1.0.5 :rocket:

PR #58, opened 2019-08-09
remove default size on icons

justinanastos:
this is not useful until we can easily override their sizes in consumers

## Release Notes

There's no easy way to override the classes that we're putting on our Space Kit components until https://github.com/apollographql/space-kit/pull/46 lands. We can't put default sizes on icons because they can't be overridden yet.


justinanastos:
:rocket: PR was released in v1.0.6 :rocket:

PR #60, opened 2019-08-09
Fix `npm watch` script

justinanastos:
It's broken because it doesn't use the `tsconfig.build.json` file.


justinanastos:
:rocket: PR was released in v1.1.0 :rocket:

PR #64, opened 2019-08-12
Automatically calculate button text color

justinanastos:
Buttons will now determine if they should use `colors.grey.darker` or `colors.white` based on how readable the text color is during the hover state. Note that this lines up perfectly with the text colors used in @adamatorres 's space kit designs!

This also removes the text color overrides in the storybook stories.


justinanastos:
:rocket: PR was released in v1.2.0 :rocket:

PR #67, opened 2019-08-13
Small text field fixes

jgzuke:
#### Allow input class names
This is needed to opt out of styling from `Input.less`

#### Make label optional again
Im not sure why this went away, adding it back since it seemed like an unintentional change to make it required.

#### Proptypes extends `InputHTMLAttributes`
Im not sure when this went away either, but without it you have to spread props to avoid type errors which isn't very ergonomic, for example
```jsx
{...{
  name: 'password',
  type: 'password',
}}
```
instead of
```jsx
name="password"
type="password"
```

### DIFF
--- a/src/TextField/TextField.tsx
+++ b/src/TextField/TextField.tsx
@@ -54,7 +54,7 @@ interface Props {
   /**
    * Visible title
    */
-  label: React.ReactNode;
+  label?: React.ReactNode;
 
   /**
    * A short hint that describes the expected value of the input field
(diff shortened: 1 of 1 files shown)


justinanastos:
I don't want to spread all the props; I want this to be a container with well-defined options. I also want to find a better solution than needing to add `.opt-out` to our `input`s.

jgzuke:
@justinanastos sounds good, I switched to just the `name` and `type` since they tend to be very common on inputs

justinanastos:
:rocket: PR was released in v1.5.0 :rocket:

justinanastos:
I just realized this PR adds the `name` prop to the component and doesn't actually use it. Yet another reason why I need to start adding tests.

jgzuke:
@justinanastos unfortunately they still got passed through in `otherProps`. Im not sure what kind of test would catch that since it was still working properly (even though it should certainly have been explicit). We could write a lint rule to say all explicitly typed props should be destructured to avoid that kind of mistake, im not sure what the use cases for exceptions to that rule would be?

PR #69, opened 2019-08-14
Correct npm script for cleaning TextField, make labels optional

mayakoneval:
## Release Notes

@justinanastos made an update to the TextField logic and didn't update the clean script. This fixes it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "clean:button": "rimraf Button",
     "clean:colors": "rimraf colors",
     "clean:fonts": "rimraf fonts",
-    "clean:formField": "rimraf FormField",
+    "clean:textField": "rimraf TextField",
     "clean:icons": "rimraf icons src/icons/*.tsx",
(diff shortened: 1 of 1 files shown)


src/TextField/TextField.tsx
```diff
@@ -49,7 +49,7 @@ interface Props {
   /**
    * Visible title
    */
-  label: React.ReactNode;
+  label?: React.ReactNode;
```
justinanastos: Now that the label is optional we need to conditionally render it as well, otherwise we'll have padding and margins on an empty element. 

I'm going to say this is good and defer that to #67 ? I'm going to 👍 both of these. If they both merge, then they'll work together 🤝

justinanastos:
@mayakoneval Remember, whatever you put in the title of a PR is what's going to be added to the changelog

justinanastos:
:rocket: PR was released in v1.5.1 :rocket:

PR #70, opened 2019-08-15
Add `as` prop to `Button`

justinanastos:
Let me start by saying I'm not thrilled about this implementation and I'm leaving it like this because I don't want to work on it anymore. 

This started as having a render function in `as` so that we could use any component we wanted and then we had some sensible props that lived on `Button`. Feedback from @cheapsteak and @trevorblades was that they didn't like how a render prop function looked (agreed, it is not as a great of an API design) and instead wanted to be able to pass a string or a `React.ComponentType` to the `as` prop and have the component magically assume all the props of those types. Turns out that's not possible. Turns out an article was written about this a few weeks ago: https://blog.andrewbran.ch/polymorphic-react-components/. Their TLDR; was don't do this, it doesn't work and will crash TypeScript 3.4; instead favor render a prop function.

I'm inclined to leave this as it is for now because I want to stop working on this (I've sunk 20 hours into it already, including too much weekend time because I was ~certain~ fooling in thinking I could make it work. If I went with the render prop method, then there's two questions to answer that I am punting for now:

1. What props do we define on `Button` so the normal use case, normal being defined as the case with `as` doesn't need to be customized, has all the props it needs.

   I think that changing things is so low friction that this isn't really a problem. If we want to add a prop, we can add it, have the PR reviewed, merged, and released in less than 10 minutes.

1. How do we type the `props` we're going to pass to the render function to prevent us from trying to spread props that a component isn't capable of consuming

    My idea here is to create a function that generates the props that we'll pass and then we can use the `ReturnType` of that function to type the props being passed to the render prop function.

---

As it stands right now, there are two ways to use `as`:

1. `keyof JSX.IntrinsicElements`, which is a string like "button" or "a". "button" is the default.

    `Button` will then accept all the props that that intrinsic element would accept. For example, using "a" will add the `href?: string | undefined` prop to `Button`, but using "button" will not allow `href`.

1. `React.ReactElement<any>`, like `<a href="apollographql.com" />` 

    The props will be spread into the element via `React.cloneElement`. `Button` will not accept any props to be spread (meaning only `Button`'s props are allowed). If you want to add extra params like `className`, add them directly to the `as` element. `classNames` are combined and `onClick` is passed through after being used internally.

There is no support for refs yet. I'm hoping we don't need them. There's also no type checking that the underlying `as` component we chose to use supports the props `Button` is passing to it.

---

All feedback is welcome, please never hold back on feedback. That being said, please consider this PR refined to the point where I don't want to make logic changes unless they are absolutely necessary because I want to get this released so we can use it so ere on the side of `NBD` changes versus "you need to change this" changes.

---

### Edit

Another thing I don't like about this approach is how not re-usable it is. We can't even re-use the `ComponentWithAs` type because it has a hardcoded default value in it. We'll have to copy-paste this and change it wherever we want the top level element to be override-able.

---

@cheapsteak discovered that if we disabled `strictFunctionTypes` and consumers of this project disabled it, then we could also enable passing `as={Component}`. We don't know if that's valuable or not so we punted the discussion. https://github.com/apollographql/space-kit/pull/70#pullrequestreview-277271643


cheapsteak:
This is a bit more verbose / boilerplatey than I'd have preferred for something that would be frequently used. 

Then again, I guess we can wrap it in the consumer app and reexport it as LinkButton, so :+1:

justinanastos:
Almost forgot @cheapsteak , this is inspired by the article you suggested https://github.com/stardust-ui/react/blob/master/MANIFESTO.md

cheapsteak:
As soon as I realized we can wrap and re-export it with the components we frequently use it with I don't think it's boilerplatey anymore.  

I might want to discourage inline ad-hoc usage of this form in the app but that can be done on a per-case/PR basis :)

justinanastos:
> As soon as I realized we can wrap and re-export it with the components we frequently use it with I don't think it's boilerplatey anymore.
> 
> I might want to discourage inline ad-hoc usage of this form in the app but that can be done on a per-case/PR basis :)
> 
> (E.g I'm thinking we would want to use this and re-export a `<NavLinkButton>` or `<LinkButton>` in engine?)

@cheapsteak I have no opinion 🤷‍♂ Creating a new component or using this inline works for me

src/Button/Button.tsx
```diff
@@ -136,11 +137,21 @@ function getHoverBackgroundColor({
...
+   *
+   * 1. A string representing the type of element you want rendered, or
+   * 2. A render function. The render function will be called with the props
+   *    that need to be passed to the element. You can use this to change the
+   *    element being rendered or to add your own props that aren't included in
+   *    this component.
+   *
+   * @default "button"
+   */
+  as?: string | ((props: any) => ReturnType<React.FC>);
```
justinanastos: Another option would be to _not_ pass `props` and instead use `React.cloneElement` to automatically add them.

That has the pro of a user can't forget and a con of it's magic. I think I like the idea of using `React.cloneElement`.

cheapsteak: Something to consider - https://twitter.com/0xca0a/status/1147099945368993793 (scroll down to see Ryan Florence weigh in)

Tangentially related, I imagine if a user forgets to pass props then it's immediately obvious that they made a mistake (since it doesn't look like a button) and should revisit the docs

Paired with if the consumer app wrapped this in a `<LinkButton>` component for reuse, there shouldn't be much opportunity for this to forgetfulness to happen

justinanastos: I'm happy you're leaning in this direction because that's where I landed as well. Trying to magically add the props falls apart when trying to use emotion.

cheapsteak:
Huh, FYI, this works 

```ts
const Button: <Props extends {}>(
  props: { as: React.ComponentType<Props> } & Props,
) => React.ReactElement = (props) => <props.as {...props} />;

// no need to <Button<NavLink>
<Button as={NavLink} to="/" />;
```

image

justinanastos:
> Huh, FYI, this works
> 
> ```ts
> const Button: <Props extends {}>(
>   props: { as: React.ComponentType<Props> } & Props,
> ) => React.ReactElement = (props) => <props.as {...props} />;
> 
> // no need to <Button<NavLink>
> <Button as={NavLink} to="/" />;
> ```
> 
> image

While this works, I really don't like the idea of dynamically adding new props to the `Button` component. What about when we want to use `partAs` (which we literally need right now so I'm opening a PR for `Input`s)

src/Button/Button.tsx
cheapsteak: @justinanastos 	let's thread this (responding to https://github.com/apollographql/space-kit/pull/70#issuecomment-521792084)

> What about when we want to use `partAs` 

What problems would be created by using `as` like this `<Button as={NavLink} to="/" />` that doesn't exist when used like this `<Button as={(props) => <Link {...props} to="some-route" />}>` wrt `partAs`?

cheapsteak: ah, is it typing? having trouble getting ts to not yell at me trying to add `partAs`

maybe more trouble than it's worth :p

cheapsteak: oh oh I got it working  

does this help?

image

justinanastos: First, like I said, I don't want to add new props to the component because we use the `as` prop. I'm trying to keep Space Kit's APIs simple; this adds this will now sometimes have a side-effect of adding ALL the props of your `as` component to the Button. That doesn't feel like a good API to me.

This also will require that we always use a non-anonymous function or rendering will unmount/mount on every render because you're using `React.createElement` (via JSX), where I was intentionally using a render function (meaning _not_ a `React.createElement`).

Going back to `partAs`, now we're going to say that "if you use `as` you use one API, and when you use `partAs` you use a different API." I'd rather keep this logic the same for `as` and `partAs`.

justinanastos: When I say a "different API", I mean props from the `<Button>` will be spread into `as`. When using `partAs` you have to call the anonymous function (which will cause a mount/unmount on every render if we're using `React.createElement`) to add props.

trevorblades:
I personally prefer the `<Button as={Link} to="/path">` API. Is the argument against it that we don't get prop type checking on the `as` component props?

Sorry if I missed it, but what is `partAs` and how would it differ from `as`?

src/Button/Button.tsx
justinanastos: @trevorblades Can you give more than "I prefer it"? Like I said here, we are intentionally limiting the number of props on the component. This will add hundreds _sometimes_ which feels like a bad API.

trevorblades: I'm not too bothered by the exposing of `as` component props on the `Button` component. The JSX reads better IMO, and as far as I can tell, doesn't impact the user negatively. In the following example, `A` looks nicer to me:

```js
// A
<Button as={Link} to="/path1" />
<Button as={Link} to="/path2" />
<Button as={Link} to="/path3" />

// B
<Button as={props => <Link to="to="/path1" {...props} />} />
<Button as={props => <Link to="to="/path2" {...props} />} />
<Button as={props => <Link to="to="/path3" {...props} />} />
```

Assuming we went with the render props method, would there be any props that are passed to the button (ie. `feel`) that we wouldn't want to pass to the underlying component (`Link` in this case)? Would I have to write my render function like this?

```js
<Button
  feel="raised"
  as={({feel, ...props}) => <Link to="/path1" {...props} />}
>
  Click me
</Button>
```

Using `React.createElement` in the `Button` component would give us control over what props are eventually passed to the underlying component.

justinanastos: If I looked at the `Button` code in example `A` then I'd assume that `to` was a prop that exists on `Button`. I'd go to the docs and I wouldn't see it and I wouldn't know why. It's that misdirection that I want to avoid.

Second, we can't use `React.createElement` because we use emotion's `css`. I tried automatically passing the props through, including the `css`, and I couldn't make it work as we'd expect.

Lastly, if we use `React.createElement` (or `jsx` provided by emotion) to render a component you pass to `as`, that means we can never use a render function because this will mount/unmount the element on every render. This is why react router provides a `render` and `component` prop on `Route`.

I think I'm going to stick with the pattern from the startdust article. Also, something I missed is that they support passing `as={<a href="" />}`. I don't know if I can support that and use `css` and have `css` allowed to be overridden. I'll check that out tomorrow.

Thanks for all the feedback and thank you @cheapsteak for getting those types to work. I'm super impressed!

cheapsteak: If it matters, the pattern I proposed comes _from_ stardust   

https://github.com/stardust-ui/react/blob/master/docs/src/pages/Composition.mdx#unhandled-props--dom-attributes

image

justinanastos: I've said it in 3 message and I'd really like you directly respond to this @cheapsteak : if `React.createElement` is called an inline function it'll mount/unmount on every render because it sees you using a different component. To my knowledge, and I tried to go this direction from the start, there's no way to know if something is an SFC or a render function. Again, why `Route` in react-router has a `Component` prop and a `render` prop that do different things (this is over a year old but explains it well, also it was the first result when I googled it https://medium.com/@ilovezcd/difference-between-render-and-component-prop-on-react-router-v4-368ca7fedbec)

justinanastos: In short, if we're going to agree we can use these forms; I don't know how to implement it:

```jsx
<Button as="button" />
```
```jsx
<Button as={<NavLink to="/"} />
```

```jsx
<Button as={NavLink} to="/" />
```

```jsx
<Button as={(props) => <NavLink {...props} to="/" />} />
```

Options 1-3 can use `React.createElement` (really emotion's `jsx` because this has to support emotion) and try to spread props. The 4th option has to be different because `React.createElement` will unmount on every render. That's not an acceptable trade-off to save 20 keystrokes when you create `<LinkButton>` component in each project that consumes this.

It also, like I keep saying, does not force us into the pit of success. We're providing an API that can very easily cause undesired and confusing behavior.

cheapsteak: What is the practical consequence of `<Button as={(props) => <NavLink {...props} to="/" />} />` unmounting on re-render? Also why would someone want to write `<Button as={(props) => <NavLink {...props} to="/" />} />` if they could have written `<Button as={NavLink} />`

This feels like a situation of letting the perfect be an enemy of the good

cheapsteak: - I believe there isn't a significant use case for `<Button as={(props) => <NavLink {...props} to="/" />} />` if we supported `<Button as={NavLink} to="/"/>`
- I believe for those situations where we need `<Button as={(props) => <NavLink {...props} to="/" />} />`, the performance cost will be negligible  
- I believe the potential (and imo minor) confusion caused by seeing a "to" prop exist on a "Button" component when it has `as={NavLink}` will not be a hindrance for the calibre of developers who are likely to work with this component library (Apollo devs)
- I believe there's value (albeit also minor) in following patterns established by styled-components, semantic-ui-react, stardust
- I believe developers who would work with this lib would prefer the ergonomics of `<Button as={NavLink} to="/"/>` (admittedly data point of 3 so far and only 2 in favour of it)


On those bases of belief, I believe `<Button as={NavLink} to="/"/>` would be a relatively more pragmatic API, but I have no way of supporting that besides intuition, and accept if your intuition differs from mine. At the end of the day, they would both solve the problem, and I think someone (I think you) will need to put their foot down :)

justinanastos: I'm fine going in that direction @cheapsteak ; I'm trying to map this to how we'll use it for `labelAs` for `<Input>`, which we also need right now.

I guess we can say "don't use a render prop function" and leave it at that.

justinanastos: @cheapsteak Can you take a shot at updating the props above to work with a default element of "button" as well?

cheapsteak: Got the typings to work with quite a bit of help from a typescript guru ( shouting out to @jephuff)

https://github.com/apollographql/space-kit/blob/53c5a73d6434ce0b60cd45acbe7fdf9486490e2e/src/Button/Button.tsx

still a bit sloppy and didn't spend much time with variable naming, but maybe you could work with this?

src/Button/Button.tsx
```diff
@@ -200,227 +197,317 @@ interface Props
   variant?: "fab";
 }
 
+/**
+ * all the types `as` can be
+ */
+type AsType = React.ReactElement | keyof JSX.IntrinsicElements;
+
+// Why isn't this in it's own file? I don't know how to make the default value
```
cheapsteak: comment still relevant?

justinanastos: No, it's not. And because it's not relevant, I can move this to another file.

justinanastos:
> LGTM
> Would you mind adding the section on the possibility of using `strictFunctionTypes` for posterity in the PR's readme?

@cheapsteak 	added.

trevorblades:
LGTM! Thanks for the detailed explanation 🙌

justinanastos:
:rocket: PR was released in v1.6.0 :rocket:

PR #71, opened 2019-08-16
update SSH fingerprint to use apollo-bot2 instead of justinanastos

apollo-bot2:


justinanastos:
:rocket: PR was released in v1.6.0 :rocket:

abernix:
@justinanastos 	Why is this additional SSH key necessary, rather than just the deployment key which is made available which already gives access to the repository? (i.e. the one owned [by CircleCI here](https://github.com/apollographql/space-kit/settings/keys) on our GitHub account and listed as an acceptable key on [this CircleCI page](https://circleci.com/gh/apollographql/space-kit/edit#checkout)).

Put another way, what would happen if the entirety of this `add_ssh_keys` object was removed?

justinanastos:
I made PR was made when I was logged in as `apollo-bot2` by accident. We should not have PRs opened by a machine-user because then we can't track who did what.

PR #73, opened 2019-08-20
fix(button): fix disabled button style

justinanastos:
Also unifies the `[disabled]` and `[disabled]:hover` styles so we don't have to repeat ourselves.


justinanastos:
I'm going to merge this because the changes are not really programatic in nature.

PR #79, opened 2019-08-22
Button: add 12px padding and expanded content example

evans:
[Torres pointed out that the buttons did not expand to fit their contents](https://github.com/mdg-private/engine-frontend/pull/1607#issuecomment-523975428), so this PR adds that padding. 

image


src/Button/Button.tsx
```diff
@@ -425,6 +425,8 @@ export const Button: ComponentWithAs<Props, "button"> = ({
                 alignItems: "center",
                 display: "flex",
                 justifyContent: "center",
+                paddingLeft: "12px",
```
justinanastos: These changes need to be applied only when buttons are not icon only. Note that the designs don't have the padding applied for icon-only buttons.

Use the [`iconOnly`](https://github.com/apollographql/space-kit/blob/master/src/Button/Button.tsx#L242) variable to conditionally add this padding.

## Design

image


## This PR

image

PR #80, opened 2019-08-23
Change `Button` `as` prop to only accept `React.ReactElement`

justinanastos:
This will unblock engine from using the latest space kit.


justinanastos:
:rocket: PR was released in v2.0.0 :rocket:

PR #87, opened 2019-08-28
Remove font weight from smaller fonts

justinanastos:


jgzuke:
Q: Out of curiosity, are we planning to add anything else to `defaultTypographyDefinitions`? It doesnt seem like its saving much over just inlining `fontFamily`

justinanastos:
> Q: Out of curiosity, are we planning to add anything else to `defaultTypographyDefinitions`? It doesnt seem like its saving much over just inlining `fontFamily`

That's a good point. There's no point in those defaults! I'll wipe them out.

justinanastos:
:rocket: PR was released in v2.2.0 :rocket:

PR #89, opened 2019-08-30
AP-783 set modal `max-height` to 60%

cheapsteak:
Since `top: 20%`, the body can be at most `height: 60%` until the top and bottom spacings are no longer even


apollo-bot2:
:rocket: PR was released in v2.2.1 :rocket:

PR #90, opened 2019-09-03
AP-714 Move emotion export docs to storybook

justinanastos:
I'm thinking we should limit how much content we have in the README. I'm thinking maybe we just include how to use the project, how to contribute to the project, and guidelines. Everything else can live in our living documentation system in storybook.

This is just the first step, we can come back to this in the future.


apollo-bot2:
:rocket: PR was released in v2.3.0 :rocket:

PR #94, opened 2019-09-06
Add `align` property to `Table`'s `columns`

justinanastos:


src/Table/Table.tsx
```diff
@@ -23,6 +24,11 @@ interface Props<RowShape> {
    * Definition of how each column will be rendered
    */
   columns: readonly {
+    /**
+     * Apply text-align all the cells in this column
+     */
+    align?: CSS.TextAlignProperty;
```
cheapsteak: I think it'd be useful to allow specifying this for the column header as well  
I think most often we'd want the alignment to match, but can't discount the possibility of needing them to differ

adamatorres: I agree, the text alignment has to do with the entire column. I can't think of a case where I'd want different alignment for the header and the content.

adamatorres: Oh, and, left align should be the default. I don't think it would be otherwise, but just wanted to state that.

justinanastos:
> No reason not to merge this to unblock us
> 
> Once we're under a bit less of a time crunch, we may want to consider standardizing how we provide values for the `th` and `td` and also making them more open for modification so things like alignment don't need API additions (I'm thinking of something similar to `react-table`'s `getTdProps` and `getTheadThProps`)

I couldn't agree more. I think now is the time since this isn't really blocking us. Maybe now is the time to use `getTdProps` or `getTheadThProps` so we don't have to deal with breaking changes later.

justinanastos:
Let's hold off on this until we think it through a little more.

PR #95, opened 2019-09-06
AP-808: passthrough `autoFocus` prop to underlying `input` in `TextField`

justinanastos:
Also adds a test to make sure it's working.

Contributes to issue AP-808

### DIFF
--- a/src/TextField/TextField.spec.tsx
+++ b/src/TextField/TextField.spec.tsx
@@ -1,0 +1,16 @@
+import React from "react";
+import { render, cleanup } from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
+import { TextField } from "./TextField";
+
+afterEach(cleanup);
+
+test("when passed `autoFocus` prop, should have focus after mounting", () => {
+  const { getByTestId } = render(
+    <TextField autoFocus inputAs={<input data-testid="input" />} />
+  );
+
+  const textField = getByTestId("input");
+
+  expect(textField).toHaveFocus();
+});
--- a/src/TextField/TextField.tsx
+++ b/src/TextField/TextField.tsx
@@ -9,6 +9,11 @@ import classnames from "classnames";
 
 interface Props {
   /**
+   * Passed through to the underlying `input`
+   */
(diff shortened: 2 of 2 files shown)


src/TextField/TextField.tsx
```diff
@@ -8,6 +8,11 @@ import { IconInfoSolid } from "../icons/IconInfoSolid";
 import classnames from "classnames";
 
 interface Props {
+  /**
+   * Passed through to the underlying `input`
+   */
+  autoFocus?: boolean;
```
jgzuke: nbd: Maybe put this down below with the other props intended for input

src/TextField/TextField.spec.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
```
jgzuke: 🎉 🎉 Tests

src/TextField/TextField.spec.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { render, cleanup } from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
+import { TextField } from "./TextField";
+
+afterEach(cleanup);
+
+test("when passed `autoFocus` prop, should have focus after mounting", () => {
+  const { getByTestId } = render(
+    <TextField autoFocus inputAs={<input data-testid="input" />} />
```
jgzuke: nbd: Maybe worth adding another `TextField` to make the test more complete?

justinanastos: I don't follow, what do you mean?

justinanastos: I'm going to merge this as-is because I need it ASAP; I'll add more tests after we talk about it :)

src/TextField/TextField.tsx
justinanastos: I'm going to merge this as-is because I need it ASAP; I'll add more tests after we talk about it :)

apollo-bot2:
:rocket: PR was released in v2.5.0 :rocket:

PR #96, opened 2019-09-07
AP-808 Add `as` prop to `Modal`

justinanastos:
This will let us control the outer element that `Modal`s are rendered with. We're doing this for AP-808 so we can make that outer container a `<form>` so we can use `onSubmit` to handle when a user presses return and when a user clicks on the `<Button type="submit">`

Contributes to issue https://apollographql.atlassian.net/browse/AP-808

### DIFF
--- a/src/Modal/Modal.tsx
+++ b/src/Modal/Modal.tsx
@@ -1,11 +1,23 @@
 /** @jsx jsx */
-import { jsx, css } from "@emotion/core";
+import { jsx, css, ClassNames } from "@emotion/core";
 import React, { useEffect } from "react";
 import * as typography from "../typography";
 import { colors } from "../colors";
 import * as CSS from "csstype";
+import classnames from "classnames";
 
 interface Props {
+  /**
+   * Override the the default element used to render the outermost component
+   *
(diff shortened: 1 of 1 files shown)


src/Modal/Modal.tsx
```diff
@@ -1,11 +1,23 @@
...
+  /**
+   * Override the the default element used to render the outermost component
+   *
+   * All props provided will be merged with props that this component adds,
+   * including `className`s being merged. You can use this to make the modal a
+   * `<form>`
+   *
+   * @default `<div />`
+   */
+  as?: React.ReactElement;
```
jgzuke: Did we want to include ` | keyof JSX.IntrinsicElements` here to be consistent?

justinanastos: Good point. We're using `as?: React.ReactElement;` everywhere we define `as`, and we're using `inputAs?: React.ReactElement | keyof JSX.IntrinsicElements;` in `TextField`.

Let's clean this up in a feature release soon.

apollo-bot2:
:rocket: PR was released in v2.6.0 :rocket:

PR #97, opened 2019-09-09
AP-801 Show appropriate cursor for `Button`'s `loading` and `disabled` states

justinanastos:
# Intent

- Loading or disabled buttons don't show a cursor pointer anymore

Contributes to issue [AP-801](https://apollographql.atlassian.net/browse/AP-801)

### DIFF
--- a/src/Button/Button.spec.tsx
+++ b/src/Button/Button.spec.tsx
@@ -64,6 +64,29 @@ test("when disabled, button does not call click handlers", () => {
   expect(rootElementOnClick).not.toHaveBeenCalled();
 });
 
+test("when loading, button does not call click handlers", () => {
+  const rootElementOnClick = jest.fn();
+  const asElementOnClick = jest.fn();
+
+  const { getByTestId } = render(
+    <Button
+      as={<button onClick={asElementOnClick} />}
+      data-testid="button"
+      loading
+      onClick={rootElementOnClick}
+    >
+      submit
+    </Button>
+  );
+
+  // act
+  getByTestId("button").click();
+
+  // assert
+  expect(asElementOnClick).not.toHaveBeenCalled();
+  expect(rootElementOnClick).not.toHaveBeenCalled();
+});
+
 test("when clicked should remove focus from itself", () => {
   const { getByTestId } = render(<Button data-testid="button">submit</Button>);
   const button = getByTestId("button");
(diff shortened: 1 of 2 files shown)


src/Button/Button.spec.tsx
```diff
@@ -64,6 +64,29 @@ test("when disabled, button does not call click handlers", () => {
   expect(rootElementOnClick).not.toHaveBeenCalled();
 });
 
+test("when loading, button does not call click handlers", () => {
```
jgzuke: FYI: Im glad we have this now 👍

src/Button/Button.tsx
```diff
@@ -353,6 +353,12 @@ export const Button: React.FC<Props> = ({
 
                 color: getTextColor({ color, feel, theme }),
 
+                cursor: loading
+                  ? "progress"
+                  : disabled
+                  ? "not-allowed"
```
jgzuke: Hmm personally I feel like we could just use `default` over `not-allowed` and `progress`. I just tried them out and they seem kind of ugly on mac at least. It might just be because not many modern websites use them, I dont think Id ever seen them before.

This is just my personal feeling so I wont request changes here, but I would be interested in hearing the reason we want them.

justinanastos: No real reason. I was concerned people would not understand they can't click on disabled buttons, but I guess I'm solving a problem we don't actually have yet. The progress one is super ugly, you're right. I'll axe these.

src/Button/Button.tsx
```diff
@@ -353,6 +353,8 @@ export const Button: React.FC<Props> = ({
 
                 color: getTextColor({ color, feel, theme }),
 
+                cursor: loading || disabled ? "default" : "pointer",
```
jgzuke: minor: This is super minor, but I always like when these are explicit about `(loading || disabled) ? "default" : "pointer"` to distinguish from `loading || (disabled ? "default" : "pointer")`. It doesnt actually make a difference but for some reason I find it a lot easier to parse.

justinanastos: Totally understand; that's actually how I wrote it but prettier had different ideas :)

justinanastos: I can't self-resolve a `minor` if I'm not making the change, so I'm requesting a re-review

image

jgzuke: Ah okay 😆This is fine!

justinanastos: Thanks! Please 👍 and we can :shipit:

apollo-bot2:
:rocket: PR was released in v2.6.1 :rocket:

PR #98, opened 2019-09-10
Document `Button as={}`

justinanastos:


src/Button/Button.stories.mdx
```diff
@@ -363,6 +363,31 @@ For dark themed buttons, the borders are completely removed to make the colors l
...
 
+## Custom element
+
+You can customize the element rendered by `Button`. This is helpful when you want to render a link that looks like a button.
+
+<Preview>
+  <Story name="as link">
+    <Button as={<a href="#" />}>Link</Button>
+  </Story>
+  <Story name="as red link">
```
jgzuke: nbd: Im not sure if changing the colors here more than once provides extra explanatory value. Have we come across any other example use cases for `as` that we could add here instead?

justinanastos: Interesting point. This isn't just trying to be explanatory, it's also to add storybook stories so we have visual validation in CI that we aren't breaking what colored buttons with `as={}` look like. I could probably add another file for just stories to test; but this felt like a more natural place for it.

apollo-bot2:
:rocket: PR was released in v2.7.0 :rocket:

PR #99, opened 2019-09-10
Add Chromatic visual testing

justinanastos:
The API key we use is stored in CircleCI, not in source control. If you want to use this locally, then you need to get the API key from chromaticqa.com and export it locally. I personally keep mine in my `.envrc` file, consumed by `direnv`.

Make this a required check when this lands.

Lastly, after this, we can enable renovate with auto merge. If anything merges that had bugs, then we need to write tests to capture those bugs. We should expect release confidence testing from this library going forward.


apollo-bot2:
:rocket: PR was released in v2.7.0 :rocket:

PR #100, opened 2019-09-10
[WIP] fix(modal): Stop modal from closing unexpectedly

justinanastos:
## Release Notes

We close modals when you click on the shroud. If you mouse down inside the modal, move the mouse outside of the modal, and then mouse-up, the modal closes. We should stop capturing `onClick` and instead look at `mouseDown`.

### DIFF
--- a/src/Modal/Modal.tsx
+++ b/src/Modal/Modal.tsx
@@ -131,7 +131,7 @@ export const Modal: React.FC<Props> = ({
     <ClassNames>
       {({ css, cx }) => {
         const propsToPass = {
-          onClick: onClose,
+          onMouseDown: onClose,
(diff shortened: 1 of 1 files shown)


cheapsteak:
Oh geeze, looked up on SO, correct solution looks like it'll be quite a hassle and more importantly adds a bit of fiddly code complexity

https://stackoverflow.com/questions/55569371/how-to-stop-propagation-on-click-when-you-release-the-mouse-outside

>The full idea is to stop listening on "click" all at once. OP need to remove any existing "click" listeners, then implement a synthetic "click-like" event by using "mousedown" and "mouseup" together.

yeesh

Wonder if anyone made a downshift for modals

PR #120, opened 2019-09-18
Increase modal max height and center contents

jgzuke:
We previously had `max-height: 60%` which was too small for a few of our modals. The problem with switching to `max-height: 80%` and `top: 10%` was that all our modals would then exist at the top of the screen, so I switched to centering the modal contents instead of setting `top`.

Fixes https://apollographql.atlassian.net/browse/AP-837 and https://apollographql.atlassian.net/browse/AP-941


apollo-bot2:
:rocket: PR was released in v2.9.0 :rocket:

PR #122, opened 2019-10-02
AP-919 Stop propagation of clicks when disabled using input as

jgzuke:
- First add a failing test for the case when using `as={<div ... >}`. The other test uses `as=` but passes a button so the disabled prop already works as intended.
- Uncomment the line that fixes the test

Fixes https://apollographql.atlassian.net/browse/AP-919

### DIFF
--- a/src/Button/Button.spec.tsx
+++ b/src/Button/Button.spec.tsx
@@ -64,6 +64,29 @@ test("when disabled, button does not call click handlers", () => {
   expect(rootElementOnClick).not.toHaveBeenCalled();
 });
 
+test("when disabled, button with 'as={}' does not call click handlers", () => {
+  const rootElementOnClick = jest.fn();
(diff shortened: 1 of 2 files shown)


src/Button/Button.spec.tsx
```diff
@@ -64,6 +64,29 @@ test("when disabled, button does not call click handlers", () => {
   expect(rootElementOnClick).not.toHaveBeenCalled();
 });
 
+test("when disabled, button with 'as={}' does not call click handlers", () => {
```
justinanastos: This needs a better description. Disabling a button when using `as={<button>}` works correctly; this needs to describe a more specific case in which it doesn't work correctly.

apollo-bot2:
:rocket: PR was released in v2.12.0 :rocket:

PR #124, opened 2019-10-02
Update auto to add Jira links

justinanastos:
Auto recently updated the jira plugin so now this should work!

https://github.com/intuit/auto/issues/567

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -4,7 +4,8 @@
   "owner": "Justin Anastos <justin@apollographql.com>",
   "plugins": [
     "npm",
-    "released"
+    "released",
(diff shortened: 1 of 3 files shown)


justinanastos:
@evans The Jira link will be added to the changelog if it's anywhere the changelog sees 😃 By default we use the PR title as the changelog line. A user can also add a `## Changelog` section to the PR description and that, instead, will be used.

https://github.com/apollographql/space-kit#changelog

apollo-bot2:
:rocket: PR was released in v2.11.0 :rocket:

PR #126, opened 2019-10-03
Add framer-motion as peer-dep; animate modal AP-812

cheapsteak:
Since framer-motion's div uses transforms for animation, I had to change the method horizontal centering from `left: 0%; transform: translate(-50%)` to `left: 0; right: 0; margin-left: auto; margin-right: auto;`

Otherwise was pretty straightforward

Next steps: update space-kit version in engine-frontend

### DIFF
--- a/src/Modal/Modal.tsx
+++ b/src/Modal/Modal.tsx
@@ -1,6 +1,7 @@
 /** @jsx jsx */
 import { jsx, css, ClassNames } from "@emotion/core";
 import React, { useEffect } from "react";
+import { motion } from "framer-motion";
 import * as typography from "../typography";
 import { colors } from "../colors";
 import * as CSS from "csstype";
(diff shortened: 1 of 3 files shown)


src/Modal/Modal.tsx
```diff
@@ -151,22 +152,35 @@ export const Modal: React.FC<Props> = ({
...
+            <motion.div
+              animate={{ opacity: 1, scale: 1 }}
+              initial={{ opacity: 0, scale: 0.9 }}
+              transition={{
+                scale: {
+                  type: "spring",
+                  stiffness: 150,
+                  damping: 200,
+                  mass: 0.2,
+                  velocity: 8,
```
justinanastos: Q: Where do these values come from?

cheapsteak: Everything but velocity comes from Adam's codesandbox :)
https://codesandbox.io/s/modal-transition-dknl9 

For some reason the velocity needs to be explicitly set to `8` otherwise the animation looks very different (not enough stretch)

justinanastos: Awesome! LGTM

apollo-bot2:
:rocket: PR was released in v2.13.0 :rocket:

PR #128, opened 2019-10-03
AP-1071 Write docs on how to use `npm link` with consuming libraries

justinanastos:
Resolves issue https://apollographql.atlassian.net/browse/AP-1071


apollo-bot2:
:rocket: PR was released in v2.14.0 :rocket:

PR #130, opened 2019-10-15
AP-786 ✨ add illustrations

justinanastos:
We have SVGs as a source and we generate the TypeScript components and the storybook stories.

Contributes to issue https://apollographql.atlassian.net/browse/AP-786


justinanastos:
Closing with a new branch name in favor of #131

PR #132, opened 2019-11-06
AP-1099 Set default size of icons to be 20px by 20px

justinanastos:


apollo-bot2:
:rocket: PR was released in v2.14.0 :rocket:

PR #134, opened 2019-11-06
AP-1099 Remove default width on icons

justinanastos:
## Release notes

We tried setting a default height and width for icons, but this ran into issues with rectangular icons where we were required to set a width and a height. Per @daniman 's suggestion we'll just set the height. I chose to set it to the viewbox height and users can override this if they want to.

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -22,9 +22,10 @@
     ],
     "patch": [
       {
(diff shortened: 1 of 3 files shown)


apollo-bot2:
:rocket: PR was released in v2.14.1 :rocket:

PR #136, opened 2019-11-11
AP-1165 Disable label checks for PRs

justinanastos:
While these are helpful for the automatic release process; we have no way
of automatically running them because GitHub actions are no longer
available on our plan as of November 13, 2019. When we find a way to
automatically run this job again, we should reenable it.

Resolves issue https://apollographql.atlassian.net/browse/AP-1165


apollo-bot2:
:rocket: PR was released in v2.14.2 :rocket:

PR #137, opened 2019-11-15
AP-1075 Fix spinner remounting restarts

jgzuke:
Take the approach outlined in https://dev.to/selbekk/how-to-stop-your-spinner-from-jumping-in-react-5cmp to stop the spinner from restarting the animation when remounted. This should fix the loader animation restarts described in https://apollographql.atlassian.net/browse/AP-1075.

This worked writing a local story that remounted the component but I was having a really rough time trying to get the linked version to run properly so 🙏 it works properly in that situation as well.

### DIFF
--- a/src/Loaders/LoadingSpinner.tsx
+++ b/src/Loaders/LoadingSpinner.tsx
@@ -73,6 +73,9 @@ export const LoadingSpinner: React.FC<Props> = ({
 
   const pixelSize = SIZE_MAP[size];
 
+  const mountTime = React.useRef(Date.now());
+  const mountDelay = -(mountTime.current % 540);
+
   return (
     <svg
       className={className}
@@ -97,6 +100,7 @@ export const LoadingSpinner: React.FC<Props> = ({
           css={{
             animation: `${SPIN} 540ms linear infinite`,
             willChange: "transform",
+            animationDelay: `${mountDelay}ms`,
           }}
           fill={asteroidColor}
           r="10"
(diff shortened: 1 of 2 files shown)


src/Loaders/LoadingSpinner.tsx
```diff
@@ -73,6 +73,9 @@ export const LoadingSpinner: React.FC<Props> = ({
 
   const pixelSize = SIZE_MAP[size];
 
+  const mountTime = React.useRef(Date.now());
+  const mountDelay = -(mountTime.current % 540);
```
justinanastos: We need to disable this for chromatic testing. Fortunately, Chromatic prescribes a way to do this (https://docs.chromaticqa.com/ignoring-elements):

```
import isChromatic from 'storybook-chromatic/isChromatic';
```

You can check the source and see that it checks the user agent, which is a trivial amount of code to add so I'm ok with it.

```js
const mountDelay = isChromatic() ? 0 : -(mountTime.current % 540);
```

justinanastos:
Rebase or merge against `master` to fix the CI failure! Sorry about that!

jgzuke:
Yep I was linking with the npm install skipped, but getting hooks errors from react version mismatches since I believe the dependencies dont get synced the same when linking. I think just pinning the versions should fix it though. Will make the chromatic change

jgzuke:
Got the link to work, it looks good in engine 👍

justinanastos:
Great job @jgzuke , LGTM!

apollo-bot2:
:rocket: PR was released in v2.16.0 :rocket:

PR #139, opened 2019-11-18
remove unnecessary scrollbar

cheapsteak:
`overflow: scroll` results in a scroll bar showing up on windows even when it doesn't make sense. should use `overflow: auto` instead

image


apollo-bot2:
:rocket: PR was released in v2.14.3 :rocket:

PR #140, opened 2019-11-18
Change loader spin duration to 750ms

jgzuke:
Change the loader spin duration from 540ms to 750ms

540ms (current) https://share.getcloudapp.com/wbue4ejW
750ms https://share.getcloudapp.com/X6u21YZX
1000ms https://share.getcloudapp.com/JruwRwk1

### DIFF
--- a/src/Loaders/LoadingSpinner.tsx
+++ b/src/Loaders/LoadingSpinner.tsx
@@ -26,53 +26,53 @@ interface Props {
   size?: Size;
 }
 
-// Length of animation
-const DURATION = 750;
-
-const SPIN = keyframes`
-  0% { transform: rotate(0) }
-  100% { transform: rotate(360deg) }
-`;
-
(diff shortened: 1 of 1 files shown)


src/Loaders/LoadingSpinner.tsx
```diff
@@ -26,6 +26,9 @@ interface Props {
   size?: Size;
 }
 
+// Length of animation
+const DURATION = 750;
```
justinanastos: NBD: I know that it's commonly held to be a good idea to store constants instead of using magic numbers; but I prefer to at least keep them in the same function so the function has fewer external dependencies, and just so I don't have to go looking around for things. Also, use jsdoc style comments so that we get hinting from vs code elsewhere (meaning `/** Length of animation */` instead of `// `)

apollo-bot2:
:rocket: PR was released in v2.17.0 :rocket:

PR #141, opened 2019-11-19
Turn off syncAnimationRestarts for chromatic and buttons

jgzuke:
Add `syncAnimationRestarts` (default true) prop to `LoadingSpinner` to replace usage of `isChromatic`. Also set to false in stories and buttons. We only really need it false in stories but this seemed easier than passing the `syncAnimationRestarts` prop to buttons as well, it also seemed like it would look more natural if button loaders weren't synced.

Im a bit conflicted about whether this should default to false instead. I feel like the default makes more sense to be false since we only really want this for our full page loaders, however it seems likely for people to miss setting the prop. This is mitigated a bit if we only use the engine wrappers instead of this component directly.

### DIFF
--- a/src/Loaders/LoadingSpinner.stories.mdx
+++ b/src/Loaders/LoadingSpinner.stories.mdx
@@ -21,19 +21,19 @@ import { colors } from "../colors";
 
 <Preview>
   <Story name="size large">
-    <LoadingSpinner size="large" syncAnimationRestarts={false} />
+    <LoadingSpinner size="large" />
   </Story>
   <Story name="size medium">
-    <LoadingSpinner size="medium" syncAnimationRestarts={false} />
+    <LoadingSpinner size="medium" />
   </Story>
   <Story name="size small">
-    <LoadingSpinner size="small" syncAnimationRestarts={false} />
+    <LoadingSpinner size="small" />
   </Story>
   <Story name="size xsmall">
-    <LoadingSpinner size="xsmall" syncAnimationRestarts={false} />
+    <LoadingSpinner size="xsmall" />
   </Story>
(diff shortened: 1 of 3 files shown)


src/Loaders/LoadingSpinner.stories.mdx
```diff
@@ -21,19 +21,19 @@ import { colors } from "../colors";
 
 <Preview>
   <Story name="size large">
-    <LoadingSpinner size="large" />
+    <LoadingSpinner size="large" syncAnimationRestarts={false} />
```
justinanastos: Q: Should all of these use `isChromatic()` instead of `false`?

jgzuke: 🤷‍♂Im not sure it matters either way, the spinner will just have a different offset when running locally

justinanastos:
You also forgot to add a label.

src/Loaders/LoadingSpinner.tsx
```diff
@@ -24,11 +23,18 @@ interface Props {
...
+   * Whether or not to sync the animation when the
+   * component is remounted, defaults to true
+   */
+  syncAnimationRestarts?: boolean;
 }
 
 export const LoadingSpinner: React.FC<Props> = ({
   theme = "light",
   size = "medium",
+  syncAnimationRestarts = true,
```
justinanastos: minor: I'm sorry I keep doing this, but please add a jsdoc comment explaining what this does so we can see that from AGM. After that, merge away!

apollo-bot2:
:rocket: PR was released in v2.19.0 :rocket:

PR #143, opened 2019-11-21
Stop stripping comments from TypeScript builds

justinanastos:
We need them for jsdoc docs to propogate to consumers!

_I'm merging this without review as soon as it passes CI._


apollo-bot2:
:rocket: PR was released in v2.17.1 :rocket:

PR #144, opened 2019-11-22
AP-1215 Make modal as prop replace inner div

jgzuke:
## Intent

Allow the inner div instead of the background to be replaced by the `as=` prop. This lets us change the style of the modal (changing things like `background-image` or `width`) while still letting us wrap modals as forms which is the current primary usage. This should fix the issue in https://apollographql.atlassian.net/browse/AP-1215 and also let us change the modal width to replace https://storybook.apollographql.com/?path=/story/engine-querytraces-querytraceinspector--default with a spacekit modal.

## How

Clone `as` to replace the inner div instead of the outermost background. The inner div is still wrapped by a `motion.div` component so we needed to change a few things so the transition still worked. This mostly meant splitting up the current styles between the `motion.div` css and `modalContents` css.

### DIFF
--- a/src/Modal/Modal.story.tsx
+++ b/src/Modal/Modal.story.tsx
@@ -1,4 +1,6 @@
+/** @jsx jsx */
 import { storiesOf } from "@storybook/react";
+import { css, jsx } from "@emotion/core";
 import { Modal } from "../Modal";
 import React, { useState } from "react";
 import { colors } from "../colors";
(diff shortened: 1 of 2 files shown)


jgzuke:
@justinanastos since the replaced div is a child of the animation div using percentage based widths won't work which kind of sucks. I didn't have a good solution for that which didn't break the animation, any ideas?

jgzuke:
Closing in favor of https://github.com/apollographql/space-kit/pull/146

PR #146, opened 2019-12-03
AP-1215 Add `as` prop to `Modal`

justinanastos:
Resolves issue https://apollographql.atlassian.net/browse/AP-1215

@jgzuke took the first stab at this in #144 and ran into an issue where wrapping our `as` element with a `motion.div` caused us to not be able to apply sizing. I had the idea to try to use `motion[as.type]` so that we can pass any `as` that we want and we'll get the correct `motion` element. It works! I had to some trickery because of of `motion.custom` accepting a generic component; we can add support for custom components when we need them. It shouldn't be tough, it's just outside the scope of this PR.

I also updated @jgzuke 's storybook story to show the subtle stripes found here: https://zpl.io/bJWMp83

# Reviewing

Look at the code commit by commit and look at the new tests.

I'd like help thinking of other tests cases for storybook stories or unit tests. I feel like there can be more confidence from additional test vectors, but I don't know what else to test 🤷‍♂

# Testing

Take a look at the new storybook story. It has diagonal strips (they are subtle)

# Other Stuff

I wanted to use an optional chaining function, so I upgraded TypeScript, which required upgrading a bunch of other stuff.


cheapsteak:
>I also updated @jgzuke 's storybook story to show the subtle stripes found here: https://zpl.io/bJWMp83

Which story has the stripes?

src/Modal/Modal.tsx
```diff
@@ -92,14 +92,25 @@ const modalBackdrop = css`
...
+
 /* istanbul ignore next */
 function assertUnreachable(value: never): never {
   throw new TypeError(`Unreachable value reached ${value}`);
 }
 
 type TLength = string | 0 | number;
 
-const getModalWidth = (size: Props["size"]): CSS.WidthProperty<TLength> => {
+function getModalWidth(size: Props["size"]): CSS.WidthProperty<TLength> {
```
jgzuke: Q: What criteria are we using here to decide function vs arrow function? I noticed you switched this but added an arrow function above. Is this just a statement vs block statement body difference?

src/Modal/Modal.spec.tsx
```diff
@@ -165,3 +167,39 @@ test("when content area is clicked, `onClose` should not be called", () => {
 
   expect(onClose).not.toHaveBeenCalled();
 });
+
+test("when passed `as` prop, props should be merged", () => {
+  const testId = faker.lorem.word();
```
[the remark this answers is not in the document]
justinanastos: To clarify: the purpose is that this is a value and we don't care what it is. Anything coming from `faker` is DNC (do not care) and everywhere else is presumably important.

I suspect the mental overhead is because we're not familiar with faker (I had to look at the typescript definitions to know what I should use), and that using it more will remove that mental overhead.

That all being said, I think I want to keep it this way.

src/Modal/Modal.tsx
justinanastos: I copied this from your PR 😃 It was failing eslint because this `const` was declared after it was used (`const` values aren't hoisted like functions). I should clean it up though.

justinanastos:
Sorry about that @cheapsteak and @jgzuke ; I didn't send the final version. I just sent it now!

justinanastos:
> > I also updated @jgzuke 's storybook story to show the subtle stripes found here: [zpl.io/bJWMp83](https://zpl.io/bJWMp83)
> 
> Which story has the stripes?

I messed up the branch while rebasing; I just pushed the final version.

src/Modal/Modal.tsx
justinanastos: I typed a message on your PR and forgot to send it talking about this point 😃

In my opinion, we should use `function`s unless we have a reason not to. An example of when we need to use a `const` would be when we have a type that defines the arguments of a function and its return type, like `React.FC`.

justinanastos:
@cheapsteak needing to have `this` be something specific is def a use case for using a `const` instead of a `function`. I think that only would be useful inside a closure; I don't think that really has any meaning for a file-level function (like this one).

justinanastos:
@jgzuke you started this work so I'm going to wait for your 👍 before merging to make sure I didn't miss anything

src/Modal/Modal.tsx
```diff
@@ -138,24 +139,30 @@ export const Modal: React.FC<Props> = ({
...
-            cx(css(modalBackdrop)),
-            as.props.className,
-            // If the parent component is using emotion with the jsx pragma, we
-            // have to get fancy and intercept the styles to use with the
-            // `ClassNames` wrapper.
-            as.props.css ? css(as.props.css.styles) : null
-          ),
-          children: (
-            <motion.div
+        return (
```
jgzuke: nbd: `{ return (` could be `(`

src/Modal/Modal.story.tsx
```diff
@@ -65,6 +67,42 @@ storiesOf("Modal", module)
...
+      as={
+        <div
+          css={css`
+            background: repeating-linear-gradient(
+              135deg,
+              ${colors.white},
+              ${colors.white} 10px,
+              rgb(255, 253, 253) 10px,
+              rgb(255, 253, 253) 20px
+            );
```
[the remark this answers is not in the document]
justinanastos: Haha thanks @cheapsteak

src/Modal/Modal.tsx
justinanastos: Totally right. Airbnb says you should always use `=> (` instead of `=> { return (`; but I've been turned off to it over time. I often want to set up a debugger or something inside and have to change the code where I shouldn't have to.

That being said, I think we should avoid changing things to mark someone's personal preference, even if that preference is mine 😃

justinanastos:
> About the function vs const, can we lint the one we prefer for top level declarations?

@jgzuke we probably could; but I think we should stick to linting things that cause errors. We'll get caught if we use a variable before it's declared; so I think we should just all write code however we prefer and agree that as long as the code isn't going to cause an issue, then it's fine. A favorite quote referenced by @cheapsteak here https://meteor.slack.com/archives/C1X798ATZ/p1566483222183800?thread_ts=1566406288.171700&cid=C1X798ATZ:

>I personally think that code uniformity is highly overrated. It's very easy to be opinionated about and loose valuable time controlling petty details.

justinanastos:
@jgzuke BTW: this is totally the opposite of my opinion of when I joined the company last year. I used to believe that we should lint everything we possibly can and make our code as uniform as possible. Now I think we should just try to solve problems and not worry about things we don't have to worry about 🤷‍♂

justinanastos:
I miss the PR check to make sure I apply a label :(

apollo-bot2:
:rocket: PR was released in v2.21.0 :rocket:

cheapsteak:
@justinanastos I was thinking `() => {}` gives you the `this` you would normally expect to have (if there were a `this` at all), but occasionally you might want to have a function where `this` is specified/bound elsewhere, like if the function expects itself to be used in a class that would be providing the `this` (like [here](https://github.com/mdg-private/monorepo/blob/master/mdg/engine/api/gateway/src/schemaRegistryService/checkSchemaForOperationBreakingChanges.ts#L27)). So I tend to segregate normal functions that doesn't expect anything weird (non-lexical) with `this` to use `() => {}`, and thus I can use `function` as a cue that maybe I need to pay attention to `this` usage

PR #147, opened 2019-12-05
Make tslib a proper dep

jgzuke:
`tslib` was added as a dev dep before, but the utilities in `tslib` are used in the compiled code which means `tslib` should actually go out as a proper part of `space-kit`. This was causing issues when using this package since `Button` was calling a spread function from tslib that wasn't found


apollo-bot2:
:rocket: PR was released in v2.21.1 :rocket:

PR #149, opened 2019-12-17
AP-1023 Use rollup to bundle library

justinanastos:
Contributes to [AP-1023](https://apollographql.atlassian.net/browse/AP-1023)

# Why

We're going to use a 3rd party tooltip library to manage our tooltips and popovers. All 3rd party tooltip libraries require some CSS be included in the final bundle. Our current bundle process does not allow this because we use `tsc`.

The solution I've chosen is to use Rollup. Rollup is intended to bundle packages together for consumption. This only builds a CJS build right now, but we can expand this to build a es module package that can be tree shaken automatically later.

Rollup does not support bundling css out of the box; we'll have to use a plugin for that (coming with the associated PR!).

# Testing

This is tested by tryin git in AGM. We should find a way to run our test suite against our compiled output files so we can guarantee that everything works. See https://github.com/mdg-private/engine-frontend/pull/1995


rollup.config.js
```diff
@@ -0,0 +1,82 @@
+/* eslint-env node */
+import typescript from "rollup-plugin-typescript2";
+import multiInput from "rollup-plugin-multi-input";
+import fs from "fs";
+import path from "path";
+
+function CJS() {
+  return {
+    input: [
+      "src/Button",
```
timbotnik: It seems strange that we'd need to modify our build config every time we add something to `src` - can we just grab everything in there?

justinanastos: You're right. I made this change. It makes the config a little more confusing, but will be future-proof. Thanks!

justinanastos: Answering your question more directly, we can't grab everything. We have to exclude certain things, but that's fine! We can exclude scripts/ directories, stories, and test files easily.

justinanastos:
I'm going to merge this in. We can roll back if necessary!

apollo-bot2:
:rocket: PR was released in v2.22.0 :rocket:

PR #150, opened 2019-12-17
AP-1023 AP-1312 Implement `Tooltip` and `ConfirmationTooltip`

justinanastos:
- Create `Tooltip` and `ConfirmationTooltip` that compose `AbstractTooltip`
- Update tests to handle new tooltips
- Update components to handle ref forwarding because tooltips need them and write tests to confirm they work
- Add tests, including testing refs, for all icons and illustrations


src/ConfirmationTooltip/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+type TippyProps = React.ComponentProps<typeof Tippy>;
+
+interface Props
+  extends Pick<TippyProps, "children" | "content">,
+    Partial<Pick<TippyProps, "placement">> {
+  /**
+   * Force the tooltip to be visible. Only use this for testing purposes; don't
+   * use this to programatically control the component
+   */
+  forceVisibleForTestingOnly?: boolean;
```
timbotnik: Is this a different take on naming scheme vs. something like `__testVisible`?

src/ConfirmationTooltip/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+        visible={forceVisibleForTestingOnly ? true : undefined}
+        {...props}
+        onShow={instance => {
+          if (timeoutRef.current) {
+            window.clearTimeout(timeoutRef.current);
+          }
+
+          timeoutRef.current = window.setTimeout(() => {
+            instance.hide();
+          }, 3000);
```
timbotnik: Seems like the delay time should be a prop?  Or "short | medium | long".  But we can also wait for a use-case.

.storybook/privatePresets/ignoreTestFilesPreset.js
```diff
@@ -0,0 +1,23 @@
+/*
+ * Ignore tests preset
+ *
+ * Modify webpack configuration to exclude all files with `.spec` or `.test` in
+ * them. We use webpack's dynamic require to generate stories for Icons; which
+ * inadvertantly causes `Icons.spec.tx` to be included. This causes storybook to
```
timbotnik: Looks like we filter out `spec` files here (should we also filter out .test files?), not quite sure how this preset ends up being needed but #notawebpackguru.  https://github.com/apollographql/space-kit/pull/150/commits/97b2cea6b42d8983099c9a7c080f376389843ea6#diff-a233c537848b6cd0d08d7c4a96487ae7R18

src/ConfirmationTooltip/index.tsx
justinanastos: I wanted to explicitly avoid configuration to guarantee consistency. I'll make sure to review this value with @adamatorres before merging. Even if we change our mine; it should be easy to quickly iterate on this.

.storybook/privatePresets/ignoreTestFilesPreset.js
justinanastos: Great question @timbotnik . This PR adds a `src/Icons/Icons.spec.tsx` and one for illustrations as well. Those use `fs` to read all the icons files and runs tests against them all. The storybook story, `Icons.story.tsx` (compiled with webpack), uses `require(` ... `)`, which will look at all possible matches and transpile them so the dynamic includes will work. This fails when trying to dynamically transpile the files using `fs`.

rollup.config.js
```diff
@@ -21,6 +21,7 @@ function CJS() {
       "src/SpaceKitProvider",
       "src/Table",
       "src/TextField",
+      "src/Tooltip",
```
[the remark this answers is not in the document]
justinanastos: Great point! I hadn't thought of dynamically including the sources; now it's automatic 😃 Thanks @timbotnik !

src/icons/scripts/convert.ts
```diff
@@ -138,40 +141,50 @@ function createCSSAttribute(css: string): types.JSXAttribute {
...
+
                   traverse(jsx, {
                     noScope: true,
                     JSXOpeningElement({ node }) {
                       updateStrokeWidths(node);
                     },
                   });
 
-                  // Add css template literal
+                  // // Add css template literal
```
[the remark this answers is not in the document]
justinanastos: Yes 🙈

.storybook/privatePresets/ignoreTestFilesPreset.js
justinanastos: Also, yes. We should also filter out .test files. Thanks 👍

src/ConfirmationTooltip/index.tsx
justinanastos: Yes. I was hoping that `ForTestingOnly` would be more obvious than `_testVisible` to those unaware of our AGM practices.

jest/setup.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable @typescript-eslint/no-empty-function */
+// eslint-disable-next-line no-undef
+global.window.document.createRange = function createRange() {
```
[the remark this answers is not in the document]
justinanastos: Great tip. Done!

apollo-bot2:
:rocket: PR was released in v2.24.0 :rocket:

src/ConfirmationTooltip/ConfirmationTooltip.spec.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <ConfirmationTooltip content={tooltipContent}>
+        <span>{interactiveElementText}</span>
+      </ConfirmationTooltip>
+    </SpaceKitProvider>
+  );
+
+  expect(queryByText(tooltipContent)).not.toBeInTheDocument();
+  userEvent.click(getByText(interactiveElementText));
+  getByText(tooltipContent);
+  jest.runTimersToTime(5000);
```
jgzuke: Should this be an exported const from the tooltips? Changing the time seems like it would break this test

src/illustrations/Illustrations.spec.tsx
```diff
@@ -0,0 +1,72 @@
+import React from "react";
+import { cleanup, render } from "@testing-library/react";
+import fs from "fs";
+// import { promisify } from "util";
```
jgzuke: Remove

justinanastos:
@jgzuke 

>Theres a lot going on in this change. I think this maybe should have been a few different prs, but it might still be worth enumerating the changes (Buttons now forward refs, Spinners forward refs, cards have headerAs, bundling changes, Icons possibly now forward refs, I cant really tell?, etc) in the description in case someone comes back to read this later, without going through the individual commit messages.
>
>As a note im a bit confused by the refs changes, I thought this was something we were trying to heavily discourage? I remember discussing adding ref forwarding to a few components but decided there were better solutions. Possibly I missed some discussions somewhere?
>
>We also seem to be using a new file organization for components where the component lives in the index file, is that expected to be the organization going forward?

You're totally right @jgzuke ; this would have been better to split into several PRs. I'll be honest; I worked on this before the Christmas break and after, and I decided I'd rather merge it in its current imperfect state and get the new tooltips into AGM ASAP instead of finishing another sprint with nothing delivered. I knew this would be a little frustrating; I decided to get this in at the cost of frustrating my teammates.

How did I make all these decisions? I developed the tooltips while replacing all the `WithTooltip` instances in AGM. I ran into having to handle icons with tooltips, buttons, and everything else. I think this landed in a pretty good place.

For each question: I only vaguely remember wanting to avoid refs; I do remember wanting to avoid spreading all props and expanding `div` props. As for the popovers, we absolutely need refs so that the library can attach DOM events.

Regarding the index file; I got frustrated that I would control-click on a file and then get brought to a type file that was just a re-export of another file, so I started leaning away from that. I'd be very happy to hear any other opinions!

src/AbstractTooltip/index.tsx
```diff
@@ -0,0 +1,57 @@
+import "../../node_modules/tippy.js/animations/shift-away.css";
```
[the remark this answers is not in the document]
justinanastos: You're right; that would be better! Originally I imported this only where we were going to use it before I created the styles component.

src/AbstractTooltip/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * Relaxed is intended for when your tooltip has a header and you need some
+   * more space
+   */
+  padding?: "normal" | "relaxed";
+}
+
+type Props = TippyProps & AbstractTooltipProps;
+
+export const AbstractTooltip: React.FC<Props> = ({
```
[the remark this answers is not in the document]
justinanastos: I got frustrated by having to make multiple clicks to get to the actual types so this felt less painful; what do you think the pros/cons are?

src/ConfirmationTooltip/ConfirmationTooltip.spec.tsx
justinanastos: I don't want to export the time because then the tests depend on an implementation detail. This does feel a little brittle; I'll reinvestigate a better way to test this.

jgzuke:
> For each question: I only vaguely remember wanting to avoid refs; I do remember wanting to avoid spreading all props and expanding div props. As for the popovers, we absolutely need refs so that the library can attach DOM events.
Whoops, I specifically meant the forwardRef changes, not the use of refs inside components.

> Regarding the index file; I got frustrated that I would control-click on a file and then get brought to a type file that was just a re-export of another file, so I started leaning away from that. I'd be very happy to hear any other opinions!
Fwiw it does work if you click through the import instead of the file, but I do agree that its annoying. If we go that direction though I dont think id be a huge fan of all my component files having the same name when im trying to navigate between them. Personally I wouldn't mind these all being top level but 🤷‍♂Maybe we could do something at build time here instead of adding the index files?

PR #153, opened 2020-01-24
AP-1335 Implement Menus

justinanastos:
Contributes to https://apollographql.atlassian.net/browse/AP-1335

# Intent

Create a menu system based on https://zpl.io/adX0ELe. Check the examples in the storybook docs to see how to compose menus together!

Uses tippy.js under the hood, just like our tooltips.

This has a bug that can't be fixed until https://github.com/atomiks/tippyjs/issues/678 lands sometime in February. Popper v1 can't reliably figure out the boundaries of a tooltip nested below a component using overflow scrolling. This is resolved in popper v2; but that's not yet included in tippy because it's a breaking change. Fortunately the project maintainers are the same people so they're working on it.

It was a challenge to make the content scrollable if it overflowed the container, but that's included.

# Tests

There aren't any. If you can think of any worth adding, I'd love to!

# Multilevel Menus

These are supported out of the box with the underlying tippy.js library; we just haven't implemented it here yet, which is fine.

# Details

This includes these new components:

* `Menu`

    The parent of a menu. Nesting doesn't work yet; we'll follow up with that if/when we need it. Tippy does support it, we just have to set it up.

* `MenuItem`

    Item in the menu

* `MenuHeader`

    Heading for a menu. Composed from `MenuItem` to inherit it's styles

* `MenuDivider`

    For divided menus

* `MenuIconSize`

    This is not exported but used internally so you can set the menu size on a `Menu` and then use context to access the desired menu size for all children.

## Release Notes

Create new `Menu` component and supporting components `MenuItem`, `MenuHeader`, and `MenuDivider`.

See https://zpl.io/adX0ELe for the original designs and the storybook docs page for usage instructions and examples.

This does not include any tests because I didn't know what to test. Any tests or ideas for tests would be most welcome!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ tsconfig.tsbuildinfo
 /SpaceKitProvider
 /Table
 /Tooltip
+/Menu*
 /typography
 
 # Generated icons stored in src directory 
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -40,6 +40,7 @@ function CJS() {
             path.join("src", "illustrations", "scripts"),
             path.join("src", "shared"),
             path.join("src", "AbstractTooltip"),
+            path.join("src", "MenuIconSize"),
           ].some(excludedPathname => filename.includes(excludedPathname))
       ),
     external: [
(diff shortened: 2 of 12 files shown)


justinanastos:
Great observation @cheapsteak ; it'll totally work with downshift!

apollo-bot2:
:rocket: PR was released in v2.25.0 :rocket:

PR #154, opened 2020-01-29
Add more font weights

trevorblades:
This branch adds Source Sans Pro bold (700) to the CSS reset's font import, as well as specifies regular (400) and bold weights for Source Code Pro. These font weights are needed for elements in the blog, which is using this CSS reset.

### DIFF
--- a/reset.less
+++ b/reset.less
@@ -1,5 +1,5 @@
 @import (inline) "node_modules/normalize.css/normalize.css";
-@import url('https://fonts.googleapis.com/css?family=Source+Code+Pro|Source+Sans+Pro:400,600');
+@import url('https://fonts.googleapis.com/css?family=Source+Code+Pro:400,700|Source+Sans+Pro:400,600,700');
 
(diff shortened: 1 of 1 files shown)


timbotnik:
@trevorblades wondering if you can capture the effect on download size?  also @adamatorres any thoughts on how we'd want to incorporate these in typography?  i think we talked about a possible future where font size and weight are 2 separate axes instead of baked into the typographies as a combined unit, but i don't quite remember where we left that.

trevorblades:
@timbotnik I tested this out, and here's what I found:

Screen Shot 2020-01-29 at 5 38 42 PM

This change downloads an extra 23.8 KB (a 40% increase) of font files. On second thought, I don't really need to make this change in Space Kit, and can just add a `link` tag to the blog and import the necessary font weights there instead.

Thanks for asking and making me reconsider! I'm gonna close this one until there's a greater need for these extra weights outside of the blog.

PR #155, opened 2020-02-06
AP-1335 Fix animations not being disabled by `SpaceKitProvider`

justinanastos:
~Animations are still happening causing flake in our tests. It appears as though setting `animation` to `undefined` causes the default of `shift-away` to be used. We can use an empty string instead to disable animations.~

The above approach is hacky. While it visually works; it is hacky and is tricking tippy.js to do something that's already natively supported. 

Instead, we should set `duration` to `0` and `updateDuration` to `0`, and then tippy.js will recognize we're not animating. This will show the same visual output as before, but this has the additional benefit of disabling animation behaviors in the codebase that tripped us up in our JSDOM tests. Now we can test that tooltips appear _and then_ disappear. Since there's no more delays for animations, now rendering and clicking things acts closer to synchronous, allowing our tests to be more succinct and faster!

Contributes to https://apollographql.atlassian.net/browse/AP-1335

### DIFF
--- a/src/AbstractTooltip/index.tsx
+++ b/src/AbstractTooltip/index.tsx
@@ -32,14 +32,17 @@ export const AbstractTooltip: React.FC<Props> = ({
   hideOnClick,
   ...props
 }) => {
-  const { disableAnimations } = useSpaceKitProvider();
+  const {
+    disableAnimations: disableAnimationsFromProvider,
+  } = useSpaceKitProvider();
+  const disableAnimations =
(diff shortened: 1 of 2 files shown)


justinanastos:
Going to reset reviews because I found a better way to disable animations that will make the library behave as one would expect in tests so we can test both menus/tooltips appearing and disappearing.

apollo-bot2:
:rocket: PR was released in v2.25.1 :rocket:

PR #156, opened 2020-02-07
AP-1335 Remove `Menu` `instanceRef` prop and add `closeOnMenuItemClick`

justinanastos:
Contributes to [AP-1335](https://apollographql.atlassian.net/browse/AP-1335)

We were exposing an `instanceRef` prop allowing `Menu` consumers to directly interact with the Popper instance. For example, allowing consumers to close the menu when a `MenuItem` was clicked.

We ended up doing this for all instances of `Menu`. @timbotnik pointed out that we'd be better off not exposing `instanceRef` somehow. Instead, let's add a config option to automatically close a `Menu` when a `MenuItem` in its `content` is clicked. 

How? It feels weird, but I didn't want to just pass callback props down to all child elements, so I created a callback context provider. This wraps `content`, not `children` (I don't know if that actually matters). We don't have nested `Menu`s yet so I haven't figured out how exactly that'll work. This also doesn't allow any `MenuItem`s to _not_ respond to click events and close the parent menu. If we start having menus with toggle switches (we probably will), then we should put some more thought into this.

## Release Notes

- Remove `instanceRef` prop on `Menu` (💥 breaking change)
- Replace `icon` with `startIcon` in `MenuItem` (💥 breaking change)
- Add `endIcon` to `MenuItem`
- Add `closeOnMenuItemClick` prop, defaulted to `true`, on `Menu`


src/MenuItem/index.tsx
```diff
@@ -64,7 +65,13 @@ interface Props
    * This element will always be rendered unless the value is `undefined`. If
    * you want an empty node, a spacer for example, use `null`
    */
-  icon?: React.ReactNode;
+  leftIcon?: React.ReactNode;
```
timbotnik: nbd: we have a precedent for `iconLeft` and `iconRight` in parts of FE codebase, maybe consistency would be preferred.

src/Menu/index.tsx
```diff
@@ -108,23 +98,46 @@ function isReferenceObject(reference: any): reference is ReferenceElement {
...
+    Parameters<
+      NonNullable<React.ComponentProps<typeof AbstractTooltip>["onCreate"]>
+    >[0]
+  >();
+  const inheritedMenuItemClickListener = useMenuItemClickListener();
   const inheritedIconSize = useMenuIconSize();
 
+  const onClick = React.useCallback<React.MouseEventHandler>(
+    event => {
+      if (!closeOnMenuItemClick) return;
```
timbotnik: If the intention is to always forward the click to a potential parent menu, shouldn't we only be making the `hide` function below execute conditionally?  E.g. get rid of the early return, and
```
  if (inheritedMenuItemClickListener) {
    inheritedMenuItemClickListener(event);
  }
  if (closeOnMenuItemClick) {
    instanceRef.current?.hide();
  }
```

src/Menu/Menu.spec.tsx
```diff
@@ -0,0 +1,93 @@
...
+        }
+      >
+        <Button>{triggerText}</Button>
+      </Menu>
+    </SpaceKitProvider>
+  );
+
+  expect(queryByText(menuItemText)).not.toBeInTheDocument();
+  userEvent.click(getByText(triggerText));
+  getByText(menuItemText);
```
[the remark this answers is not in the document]
justinanastos: `getByText` intrinsically is an assertion. `getBy*` methods throw if they don't synchronously find what you're looking for. Adding `expect(getByText()).toBeInTheDocument()`, while it's more what we're used to, is superfluous.

I had this conversation with @jgzuke too and asked him to trust me that this will become second nature. If we find that this burns us somehow in the future, then let's revisit it.

src/Menu/index.tsx
justinanastos: Yes, this would totally bite us for nested menus. Good catch.

src/MenuItem/index.tsx
justinanastos: I can't put my finger on _why_, but I prefer it to sound conversational, left icon versus icon on the left. Also, after looking at the material ui naming conventions, I like the idea of copying flexbox's naming conventions and using `startIcon` and `endIcon`.

src/Menu/index.tsx
justinanastos: Done!

src/MenuItem/index.tsx
justinanastos: Renamed.

apollo-bot2:
:rocket: PR was released in v3.0.0 :rocket:

src/Menu/Menu.spec.tsx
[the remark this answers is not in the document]
timbotnik: gotcha, thx for the explanation.  it _seems_ like an inconsistency and not quite intuitive though.  will leave it up to you.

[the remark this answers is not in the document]
justinanastos: @timbotnik You're totally right that it seems inconsistent and not intuitive. I'm weighing this against the other option:

```js
expect(getByText(menuItemText)).toBeInTheDocument()
```

If `getByText` fails, then the code with `throw` and the `expect` will never get checked (so the error message won't actually be something related to being in the document). If it returns, then the expect will _always_ pass. 

I prefer the concise version. If we find this causes us problems then we should talk about it :-D

PR #157, opened 2020-02-07
AP-1335 Fix broken animations again

justinanastos:
I didn't push a commit the last time :( Also, I tried to write tests for this but didn't have much luck. I checked this one manually.

### DIFF
--- a/src/AbstractTooltip/index.tsx
+++ b/src/AbstractTooltip/index.tsx
@@ -1,6 +1,6 @@
 import "../../node_modules/tippy.js/animations/shift-away.css";
 import React from "react";
-import Tippy from "@tippy.js/react";
+import Tippy, { tippy } from "@tippy.js/react";
 import { TippyStyles } from "./abstractTooltip/TippyStyles";
 import { useSpaceKitProvider } from "../SpaceKitProvider";
 import classnames from "classnames";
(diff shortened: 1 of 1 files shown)


src/AbstractTooltip/index.tsx
```diff
@@ -52,8 +52,14 @@ export const AbstractTooltip: React.FC<Props> = ({
         className={classnames(className, {
           "space-kit-relaxed": padding === "relaxed",
         })}
-        duration={disableAnimations ? 0 : props.duration}
-        updateDuration={disableAnimations ? 0 : props.updateDuration}
+        duration={
+          (disableAnimations ? 0 : props.duration) ??
```
jgzuke: I think the () might be more clear wrapping `(props.duration ?? tippy.defaultProps.duration)` since thats what were actually falling through

justinanastos: Good idea! I'll do that

justinanastos: When I did that eslint got rid of it because then I'm no longer overriding default behavior 🤷‍♂

justinanastos: That's still better though, even if we have to look up order of operations with `??`.

apollo-bot2:
:rocket: PR was released in v2.25.2 :rocket:

PR #158, opened 2020-02-13
AP-1335 Fix `MenuItem`s not being truncated properly

justinanastos:
This was the result of a lot of tinkering. This makes it possible to truncate menu items 🤷‍♂


justinanastos:
@jgzuke 

> Weird, maybe a story so it doesnt break later without anyone noticing

Thanks for calling me out; I was being lazy. Story added!

src/ConfirmationTooltip/ConfirmationTooltip.story.tsx
```diff
@@ -1,21 +1,33 @@
...
-    return (
-      <div style={{ padding: 50 }}>
-        <ConfirmationTooltip
-          content="default content"
-          forceVisibleForTestingOnly
+  .add(
+    "forced visible",
+    () => {
+      return (
+        <PerformUserInteraction
```
jgzuke: Interesting, will these always trigger the new state before the snapshot is taken? I like this a lot better than testing only props. Would love to use this pattern in agm as well

justinanastos: This will be incorporated into AGM in https://github.com/mdg-private/engine-frontend/pull/2016.

While I think this is a step in the right direction, it is not an ideal solution still. Chromatic does not have an async rendering feature (https://docs.chromaticqa.com/resource-loading#asynchronous-rendering), so we have to manually add `{ chromatic: { delay: number } }` as the second argument to `.add`. @timbotnik suggested I write a blog post on this and gently nudge the chromatic team to offer the ability to programmatically delay snapshots until something has happened, which I think we be awesome.

All that being said, I'm ok adding a 2 second delay to a chromatic capture since they all run in parallel and this would enable us to write less magic props to force component states.

justinanastos: To answer your question more succinctly 

> will these always trigger the new state before the snapshot is taken

No, it will not. We have to manually delay the snapshot with a best-guess time.

apollo-bot2:
:rocket: PR was released in v3.0.1 :rocket:

src/ConfirmationTooltip/ConfirmationTooltip.story.tsx
jgzuke: Ah okay, that makes more sense. I still really like this setup though! Wasnt a huge fan of test props for stories

PR #160, opened 2020-03-11
📚 Table: expand `colProps` docs to explain applicable styles

justinanastos:
`colProps` can affect a very specific and small number of properties;
document them so we don't forget.

Note that this is labeled as "bug fix" only because I want this to be a patch bump because I want these docs to be included in VS Code.


apollo-bot2:
:rocket: PR was released in v3.1.1 :rocket:

PR #162, opened 2020-04-01
Add `containerAs` prop to Modal

justinanastos:
I need to apply a z-index to the outer element, this has been the standard
way to do that.

This is what I'm trying to fix by adding a `z-index`:

image

### DIFF
--- a/src/Modal/Modal.spec.tsx
+++ b/src/Modal/Modal.spec.tsx
@@ -189,3 +189,26 @@ test("when passed `as` prop, props should be merged", () => {
   expect(getByTestId(testId).localName).toBe("form");
   expect(getByTestId(testId)).toHaveStyle("font-weight: bold");
 });
+
+test("when passed `containerAs` prop, props should be merged", () => {
+  const testId = faker.lorem.word();
+  const className = faker.lorem.word();
+
+  const { getByTestId } = render(
+    <Modal
+      containerAs={
+        <span
+          className={className}
+          data-testid={testId}
+          style={{ fontWeight: "bold" }}
+        />
+      }
+      size="small"
+      title={faker.lorem.word()}
+    />
+  );
+
+  expect(getByTestId(testId)).toHaveClass(className);
+  expect(getByTestId(testId).localName).toBe("span");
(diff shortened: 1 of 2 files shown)


justinanastos:
Good question @cheapsteak . This isn't the backdrop, it's the container that wraps the content and the backdrop.

justinanastos:
@cheapsteak I added a screen capture of what I'm trying to correct with this change.

justinanastos:
As you can see, the backdrop _and_ the content are covered. This will allow us to modify the top level div. The `as` prop said that's what it was targeting, but it wasn't.

I'd much rather have changed `as` to `contentAs` and then used `as` for the top-level div, but that's a breaking change I'm not ready to make.

cheapsteak:
If we're planning to make breaking changes, I'd love if we could consider exporting more composable options, even if we also export pre-composed alternatives as well. I'm thinking of something that looks like this

```
<Modal.Backdrop>
    <Modal.Container>
        <Card>...</Card>
    </Modal.Container>
</Modal.Backdrop>
```

I think this would be a lot simpler to override, so that instead of `xAs` `yAx` `xProps` `yProps`, `xClassName` `yClassName` that feel almost like a DSL and option-bloaty, we'd have building blocks with very similar looking / intuitive props

justinanastos:
> If we're planning to make breaking changes, I'd love if we could consider exporting more composable options, even if we also export pre-composed alternatives as well. I'm thinking of something that looks like this
> 
> ```
> <Modal.Backdrop>
>     <Modal.Container>
>         <Card>...</Card>
>     </Modal.Container>
> </Modal.Backdrop>
> ```
> 
> I think this would be a lot simpler to override, so that instead of `xAs` `yAx` `xProps` `yProps`, `xClassName` `yClassName` that feel almost like a DSL and option-bloaty, we'd have building blocks with very similar looking / intuitive props

I'd be super happy to have this conversation @cheapsteak . I haven't seen the value in that so I'd like to talk about so I can try to understand.

apollo-bot2:
:rocket: PR was released in v3.2.0 :rocket:

PR #165, opened 2020-04-07
Restore CI check to require a label be applied before merging a PR

justinanastos:
I removed these because GitHub told us we had to. @abernix [suggested](https://meteor.slack.com/archives/C1X798ATZ/p1586262468024900?thread_ts=1585579478.009300&cid=C1X798ATZ) that I check to see if they really don't work anymore. Turns out they. Turning this back on 😃

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -2,37 +2,45 @@
   "name": "Apollo GraphQL",
   "email": "opensource@apollographql.com",
   "owner": "Justin Anastos <justin@apollographql.com>",
+  "onlyPublishWithReleaseLabel": true,
   "plugins": [
     "npm",
     "released",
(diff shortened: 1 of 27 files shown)


abernix:
For what it's worth, we're not technically paying to use it on our plan, but we're grandfathered into using GitHub Actions on our plan because we requested early beta access to it.

We may eventually move to a plan that supports this natively (and I've made some attempts to make this achievable by slowly changing organization members to external collaborators to eliminate the 4x increase in cost that we'd otherwise incur), but I'm not concerned about it going away right now, so you should feel comfortable using this!

apollo-bot2:

PR #166, opened 2020-04-15
AR-1541 Upgrade underlying dependencies and logic in menus and tooltips

justinanastos:
Contributes to [AR-1541](https://apollographql.atlassian.net/browse/AR-1541)

This was a tough endeavor. I took the code that powers popper's flip functionality ([source](https://github.com/popperjs/popper-core/blob/master/src/modifiers/flip.js)) and enhances it to allow a few fancy things:

* This enables automatic flipping and scrolling, which we couldn't do before
* Enables the scrolling to be inside the `.tippy-box` child of the root popper so the border will not be scrolled, just the content. This is why I didn't open source this because that part was super custom.
* Stop thrashing with open menu stories by disabling animations out of the box

The code is challenging and untested. I'd love to spend some time figuring out how to test this, but the best way to test this now is some new storybook stories that show the placement is where we'd expect it to be.

# Testing

Test all the menus in https://github.com/mdg-private/engine-frontend/compare/justin/space-kit/AR-1541/upgrade-menus?expand=1, specifically the service/variant selections in the left nav and on the metrics page to filter by operations.

## Release Notes

* Upgrade `@tippyjs/react` library to v4

    This includes the new popper v2, greatly simplifying how we modify the default behavior of popups, tooltips, and menus

* Upgrade Menu flip logic to attempt to flip the menu to somewhere it'll fix on the window, and if it doesn't, to pick the placement to show the maximum amount of content while enabling scrolling

* Animations can now be disabled out of the box

    Hopefully this will stop our Chromatic story thrashing. We'll see 🤷‍♂️ 

* [Breaking change] Remove `flip` and `scrollableContent` props, they are always enabled now


src/Menu/menu/sizeModifier.ts
```diff
@@ -0,0 +1,292 @@
...
+}
+
+/**
+ * Padding can be applied to `preventOverflow`, `flip`, or this modifier. Go
+ * through all the modifier options to find it
+ */
+function getPaddingFromState(
+  state: ModifierArguments<Options>["state"]
+): Padding {
+  return state.orderedModifiers.reduce<Padding>((accumulator, modifier) => {
```
jgzuke, line 126: This seems like a fancy find? `accumulator` made me assume a sum was happening, would result or something be a better name?

src/Menu/menu/sizeModifier.ts
```diff
@@ -0,0 +1,292 @@
...
+      padding,
+    });
+
+    let mainVariationSide: BasePlacement = isVertical
+      ? isStartVariation
+        ? "right"
+        : "left"
+      : isStartVariation
+      ? "bottom"
+      : "top";
```
jgzuke, line 231: Prettier is wack sometimes 😕

src/Menu/menu/sizeModifier.ts
```diff
@@ -0,0 +1,292 @@
...
+
+    if (referenceRect[len] > popperRect[len]) {
+      mainVariationSide = (getOppositePlacement(
+        mainVariationSide
+      ) as any) as BasePlacement;
```
jgzuke, line 236: You shouldnt need the as any here, afaict `Placement` should be assignable to `BasePlacement`?

src/Menu/menu/sizeModifier.ts
```diff
@@ -0,0 +1,292 @@
...
+    return {
+      overflowDict,
+      overflowPixels: Math.max(0, overflowDict.overflow[heightProp]),
+      // The `overflow` can be negative here; this is how the container can
+      // expand when we are resizing to expose more of the clipped content.
+      maxHeight: height - overflowDict.overflow[heightProp] - y,
+    };
+  });
+
+  overflowsWithOverflowMap.sort((a, b) => {
```
jgzuke, line 92: Can this be a reduce that just keeps the overflow with the smallest placement weight?

src/Menu/menu/sizeModifier.ts
```diff
@@ -0,0 +1,292 @@
...
+    );
+  }, []);
+
+  const referenceRect = state.rects.reference;
+  const popperRect = state.rects.popper;
+
+  /**
+   * Array calculated from `placements` with the calculated values of each
+   * `placement`. This will be used to determine if we're capable of flipping
+   * the element to display it ir of we have to set the `max-height` too.
```
jgzuke, line 209: or if

jgzuke, line 209: Is this comment still accurate? This doesnt seem like what we really end up using this for?

src/Menu/menu/sizeModifier.ts
cheapsteak: wish we could use lodash here 
I think this could replace the sort?
```
_.minBy(overflowsWithOverflowMap, (overflow) => overflow. overflowPixels)
```

src/Menu/menu/sizeModifier.ts
cheapsteak: how do folks feel about 
```ts
state.orderedModifiers.flatMap(
  (modifier) => modifier.options?.padding || []
)[0] || 0;

```

src/Menu/menu/sizeModifier.ts
```diff
@@ -0,0 +1,292 @@
...
+}: {
+  overflowDicts: readonly OverflowDict[];
+  state: ModifierArguments<Options>["state"];
+}) {
+  const overflowsWithOverflowMap = overflowDicts.map((overflowDict) => {
+    const { y } = state.modifiersData.preventOverflow || { y: 0 };
+    const { height } = state.rects.popper;
+    const [basePlacement] = overflowDict.placement.split("-");
+
+    const heightProp = basePlacement === "top" ? "top" : "bottom";
```
cheapsteak, line 81: I don't quite understand this variable name

Since this seems to only be used to reference overflow, how about
```ts
const verticalOverflow = overflowDict.overflow[basePlacement === "top" ? "top" : "bottom"]
```

src/Menu/menu/sizeModifier.ts
```diff
@@ -0,0 +1,292 @@
...
+
+    return {
+      overflowDict,
+      overflowPixels: Math.max(0, overflowDict.overflow[heightProp]),
```
cheapsteak, line 85: Is this assuming we'll only have vertical scroll?
Now that I just typed that out, why do we support any kind of scrolling in tooltips? Seems like a very odd interaction

jgzuke: This isnt just tooltips right, also menus?

cheapsteak: ah yup, I should have just looked at the filename 😅

src/Menu/menu/sizeModifier.ts
justinanastos: This _is_ totally a fancy find. A comment explaining this would help. I need to find the _last_ item in `orderedModifers` that defines `padding` and then default to 0 if one isn't found.

src/Menu/menu/sizeModifier.ts
justinanastos: `Placement` is not assignable to `BasePlacement`, `BasePlacement` is assignable to `Placement`. In other words, `BasePlacement` is a subset of `Placement`. There's no way for `getOppositePlacement` to know its input is a `BasePlacement` and, therefore, its output would be the same. It would make sense to add a generic to `getOppositePlacement`, but I opted instead fo create `getOppositeBasePlacement` so there's no confusion or typing issues.

src/Menu/menu/sizeModifier.ts
justinanastos: I think this sort is the best implementation to solve the problem. I've added a comment explaining the sort reasoning. We need to find the smallest `overflowPixels` and ties must go to the first `placement`, and sort doesn't guarantee the items that match the comparitor will retain their original order.

src/Menu/menu/sizeModifier.ts
justinanastos: Updated the comment; it was still accurate but the logic was convoluted.

src/Menu/menu/sizeModifier.ts
```diff
@@ -0,0 +1,292 @@
...
+    state,
+  });
+
+  return {
+    placement: minimumOverflowPlacement.overflowDict.placement,
+    maxSize: { height: minimumOverflowPlacement.maxHeight },
+  };
+}
+
+export const sizeModifier: Modifier<Options> = {
```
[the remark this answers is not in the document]
justinanastos: Yes, absolutely. Comment added 👍

src/Menu/menu/sizeModifier.ts
```diff
@@ -0,0 +1,292 @@
...
+  name: "maxSize",
+  enabled: true,
+  phase: "main",
+  requiresIfExists: ["offset", "preventOverflow", "flip"],
+  data: { _skip: false },
```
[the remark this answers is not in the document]
justinanastos: Comment added documenting use of `_skip`. I stole the logic from popper's flip implementation, but that's no excuse to not explain it to those who haven't looked through popper's source code 😃

src/Menu/menu/sizeModifier.ts
```diff
@@ -0,0 +1,292 @@
...
+    // placement, then change it.
+    if (
+      modifierArguments.state.placement !== placement &&
+      !modifierArguments.state.modifiersData[modifierArguments.name]._skip
+    ) {
+      modifierArguments.state.modifiersData[
+        modifierArguments.name
+      ]._skip = true;
+      modifierArguments.state.placement = placement;
+      modifierArguments.state.reset = true;
```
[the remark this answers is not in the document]
justinanastos: I added a comment and a link to the docs 👍

src/Menu/menu/sizeModifier.ts
justinanastos: I [stole](https://github.com/atomiks/popper.js/blob/59428b21e4081a80da44c6641a9688e6a862cb7a/src/modifiers/maxSize.js#L17) the code here so I don't know why it's named the way it is. I'm happy to change it to `verticalOverflow`.

src/Menu/menu/sizeModifier.ts
justinanastos: Right, this is _only_ menus, not tooltips.

src/Menu/menu/sizeModifier.ts
jgzuke: Sorry I meant that the other way around. I agree you still need the `as BasePlacement` here, but not the `as any`

babel.config.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-env node */
```
jgzuke, line 1: Im confused, this file existed before right? Is there a reason this is different from our babelrc?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@apollo/space-kit",
-  "version": "3.4.0",
+  "version": "3.4.1-canary.0e047a6.0",
```
jgzuke, line 3: Can we bump this to a minor version?

justinanastos: Ewwwww this should not have been in the PR!

babel.config.js
justinanastos: This is a new file for space-kit. We had to do this previously in engine-frontend.

package.json
justinanastos: Removed; the version bump happens _after_ merge; this was an accident 😃

apollo-bot2:

PR #168, opened 2020-04-17
adds another operations icon 

caydie-tran:
adding this icon
image


justinanastos:
This adds an operations icon named `src/icons/svgs/navigation/icon-operations.svg` when we already have `src/icons/svgs/navigation/icon-operations-sl.svg`. I never planned on having an icon with the same name as `-sl`; so this causes the library to get confused because when it converts `svg` files to TypeScript files it uses the filename to make the new filename and strips off `-sl`. So `icon-operations.svg` and `icon-operations-sl.svg` will both be read and will output the same filename, causing an overwrite. Then storybook will look at the source `.svg` files because it needs to know if an svg is a streamline icon or not, which is why we see the same icon on there twice.

TLDR; Delete `src/icons/svgs/navigation/icon-operations-sl.svg` in this PR and you'll be all set.

I should add a test that verifies that there aren't duplicates so that we know exactly why things are behaving strangely strange.

apollo-bot2:
:rocket: PR was released in [`v4.1.0`](https://github.com/apollographql/space-kit/releases/tag/v4.1.0) :rocket:

PR #169, opened 2020-04-20
Fix broken release process

justinanastos:
I changed the auto release configuration to include the option `onlyPublishWithReleaseLabel`, which prevents releases from being executed if there isn't a `release` label on the PR. This should fix that. 

Pushing through without a review to get this rolling.


apollo-bot2:

PR #170, opened 2020-04-22
Add success solid icon

jgzuke:
Add icon-success-solid from streamline


jgzuke:
@cheapsteak streamline is the icon library we use https://app.streamlineicons.com/

PR #171, opened 2020-05-01
Add className to Tooltip prop types

Jephuff:
Tooltip passes all props to AbstractTooltip already, so this just updates Tooltips props type to accept className.


apollo-bot2:
:rocket: PR was released in [`v6.0.2`](https://github.com/apollographql/space-kit/releases/tag/v6.0.2) :rocket:

justinanastos:
@Jephuff Why did we need `classname` on tooltip?

Jephuff:
> @Jephuff Why did we need `classname` on tooltip?

I don't entirely remember, I needed to add some tailwind styles to remove an inherited style, maybe text-transform? 
That usage has since been removed though :)

PR #174, opened 2020-05-11
AR-572 Create Alert component

Jephuff:
https://apollographql.atlassian.net/browse/AP-572

Create an `Alert`


src/Alert/Alert.tsx
```diff
@@ -0,0 +1,106 @@
...
+  children,
+  actions,
+  ...cardProps
+}) => {
+  const { Icon, color } = useMemo(() => {
+    switch (level) {
+      case "info":
+        return { color: colors.blue, Icon: IconInfoSolid };
+      case "warn":
+        return { color: colors.yellow, Icon: IconWarningSolid };
```
jglovier, line 37: The warning icon should be using `orange-base`/`F59140`.

src/Alert/Alert.tsx
```diff
@@ -0,0 +1,106 @@
+/** @jsx jsx */
+import { jsx, css } from "@emotion/core";
+import React, { useMemo } from "react";
+import PropTypes from "prop-types";
+
+import { Card, CardSeperator, CardProps } from "../Card";
+import { IconErrorSolid } from "../icons/IconErrorSolid";
+import { IconInfoSolid } from "../icons/IconInfoSolid";
+import { IconWarningSolid } from "../icons/IconWarningSolid";
```
jglovier, line 9: Also the warning icon itself should be using `orange-base` as the triangle bg color.

jglovier:
Also, I'm not sure how this should be implemented (whether via the SpaceKit component or on a per view basis), but the width of these cards should constrained to a max-width.

The [Alert inline banners](https://apollographql.quip.com/AgFcAwybcEWN/Alerts-design-principles#BTJACAw2L82) should flow with the content they are in, but toast cards will sit above the UI and slide in from off screen (sorry, that probably wasn't very clear from the design principles doc! 🙈) and generally should be constrained to a max-width in real use cases.

justinanastos:
@AlexanderMann if you're cool working on this, definitely give it a shot. I'd start with a fresh PR and _not_ extend from `Card`. Just style this on its own. We can decide soon if we want to abstract out the pattern where we have a rounded box with a shadow and background; that's not necessary here though, just copy-paste.

As for the colors, please accept a `color: PaletteColor` prop (`color` might be misleading, maybe `accentColor`?) and decide if you want that color to be for the text or the icon. Then use [`getOffsetInPalette`](https://github.com/apollographql/space-kit/blob/b3fa01c71208ae0e1f0e15b32e85cf846b80a142/src/colors/utils/getOffsetInPalette.ts#L24-L41) to get a darker or lighter shade that you'd need for the icon/text.

I'd say give the component a `title` prop and then have the content rendered with `children`. I think that makes the props look like this:

```js
interface AlertProps {
  /**
   * Color to apply to the (you choose, icon or text). We will programmatically
   * pick a darker/lighter shade for the other one (and that will presumably
   * change for a light and dark theme!) 
   */
  accentColor: PaletteColor;
  title: React.ReactNode;
}
```

I'm going to be off for the long weekend tomorrow but I'm happy to pop on and look at this if it's blocking.

cheapsteak:
> if you're cool working on this, definitely give it a shot. I'd start with a fresh PR and not extend from Card.

I have on occasion wished for a styled-component that looked like Card but didn't come with all the dressings that Card does, possibly a `BaseCard` or `CardBase`? food for thought

Jephuff:
Updated the alert to not extend card, accept color as prop and accept icon as prop.
I also added a dark theme.
image

@cheapsteak I looked at making a common CardBase component, not a while lot is actually common between the 2
```
<section
    {...otherProps}
    css={{
      backgroundColor: colors.white,
      color: colors.black.base,
      boxShadow: `0 4px 8px 0 rgba(0, 0, 0, .04)`,
      borderStyle: "solid",
      borderWidth: 1,
      borderColor: colors.silver.dark,
    }}
  >{children}</section>
```
I'm going to copy-paste for now :D

cheapsteak:
> @cheapsteak I looked at making a common CardBase component, not a whole lot is actually common between the 2
> 
> ```
> <section
>     {...otherProps}
>     css={{
>       backgroundColor: colors.white,
>       color: colors.black.base,
>       boxShadow: `0 4px 8px 0 rgba(0, 0, 0, .04)`,
>       borderStyle: "solid",
>       borderWidth: 1,
>       borderColor: colors.silver.dark,
>     }}
>   >{children}</section>
> ```

I was hoping we'd be able to pull a border radius as well, but other than that that's pretty much what I was after  

I wonder if you've caught this thread https://meteor.slack.com/archives/CJM405HJB/p1588779273039400 ?  
Pulling in the Card component and overriding its overflow by `& > div > div { ... }` feels unpleasant when all he needed were borders and shadows  

I would have just inlined the shadows myself, but my understanding is that we are heavily encouraging if not mandating that space kit be used whenever possible. That behooves us to try to provide a pleasant developer experience, and in my experience it's more pleasant to use something that gets me 80% of the way there even if I have to add the rest myself, than to have something that gets me 120% of the way there and I have to override the unwanted extras. Although, I guess this might be subjective?

Anyway, feel free to punt that from this PR as we can fairly say this is out of scope :)

Jephuff:
> Also, I'm not sure how this should be implemented (whether via the SpaceKit component or on a per view basis), but the width of these cards should constrained to a max-width.
> 
> The [Alert inline banners](https://apollographql.quip.com/AgFcAwybcEWN/Alerts-design-principles#BTJACAw2L82) should flow with the content they are in, but toast cards will sit above the UI and slide in from off screen (sorry, that probably wasn't very clear from the design principles doc! 🙈) and generally should be constrained to a max-width in real use cases.

My goal with this was to provide the UI to be used in different places like a Toast component (or modal, or inline). so I think max-width and if the card flows with the rest of the document would be determined by the usage 👍

jglovier:
>My goal with this was to provide the UI to be used in different places like a Toast component (or modal, or inline). so I think max-width and if the card flows with the rest of the document would be determined by the usage

Just to clarify, this design is only intended for use as a toast message. By definition ([Bootstrap example](https://getbootstrap.com/docs/4.2/components/toasts/), [Primer example](https://primer.style/css/components/toasts), [Android example](https://developer.android.com/guide/topics/ui/notifiers/toasts)) a toast is a temporary message that doesn't belong in the normal flow of UI and generally is triggered by an action the user has taken during their workflow, generally appearing from off-screen or floated into the view from an hidden state.

This component should never be used as a modal or inline alert.

The use case for inline alert banners differs from the toast message (which is why the design is styled explicitly different from toasts). Those should be displayed when something in the app, data, or user's account is in a state that requires an alert (vs being triggered by an immediate action). 

For example:
- deprecated fields will always show an inline warning alert next to them in the UI
- when we are changing a feature and it may go away, we might show an inline alert in the UI until the feature is gone
- if there is a problem with your account (like payment method missing) we might show an inline alert on the account page until it's resolved

Point being these will stay in the layout permanently until the issue is resolved, verses a toast message which appears momentarily until you dismiss it or reload the page.

Apologies if all of this wasn't very clear in the quip doc. 🙏 

All that to say, keeping toast cards with a consistent width (and displaying them in the same place) will be an important part of their usage in the app. If the best thing to do is to implement that on a case by case basis (rather than in the component), that's cool. I just want to make sure the context for how these will be used is clear. :smile:

Jephuff:
For anyone following along
@jglovier and I had a sync conversation, I had misinterpreted a change as getting rid of the extended/not extended versions so I'll add that change back.
And we will keep this outside a toast and add that functionality later :)

cheapsteak:
Glad we did a one-off for studio first so we can take the time to get this right 😄

src/Alert/Alert.tsx
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * The color used for the heading icon, the heading text will be 2 shades
+   * darker if light theme and 2 shaders lighter if a dark theme
+   */
+  color: PaletteColor;
+
+  /**
+   * The icon displayed to the left of the heading text
+   */
+  Icon: React.ComponentType<{ className?: string }>;
```
justinanastos, line 28: I'd like to change this to be `React.ReactElement<{ className?: string }>`. 

If we pass an element, then we can render any element with whatever props we want and then decorate it with this component much the same way we decorate all the `as` props. This allows us to configure anything else we wanted on the element. If we go the route of using `ComponentType`, then we have to assume the only configurable prop we'd pass in is `className` or we'd have to do some trickery to allow us to pass other props in. For example, if we used a Space Kit icon and we wanted to use `weight="thin"`, that's not possible if we use `React.ComponentType`, but it's trivial if we use an element: `icon={<SomeIcon weight="thin" />}`.

<details>
<summary>Here's the patch to apply</summary>

I'm sure you already know this, but as a refresher, copy this including a trailing line break and use `pbpaste | git apply -p1` to apply this patch yourself.

```diff
diff --git a/src/Alert/Alert.story.mdx b/src/Alert/Alert.story.mdx
index 13cc113..f9714d9 100644
--- a/src/Alert/Alert.story.mdx
+++ b/src/Alert/Alert.story.mdx
@@ -31,7 +31,7 @@ Informational alerts present general info about something you are doing or worki
     <Alert
       onClose={() => undefined}
       color={colors.blue.base}
-      Icon={IconInfoSolid}
+      icon={<IconInfoSolid />}
       heading="Informational alert"
     >
       This is a general informational message telling you about something. Learn
@@ -42,7 +42,7 @@ Informational alerts present general info about something you are doing or worki
     <Alert
       onClose={() => undefined}
       color={colors.blue.base}
-      Icon={IconInfoSolid}
+      icon={<IconInfoSolid />}
       theme="dark"
       heading="Informational alert"
     >
@@ -59,7 +59,7 @@ Informational alerts present general info about something you are doing or worki
     <Alert
       onClose={() => undefined}
       color={colors.blue.base}
-      Icon={IconInfoSolid}
+      icon={<IconInfoSolid />}
       heading="Informational alert"
       extended={true}
     >
@@ -71,7 +71,7 @@ Informational alerts present general info about something you are doing or worki
     <Alert
       onClose={() => undefined}
       color={colors.blue.base}
-      Icon={IconInfoSolid}
+      icon={<IconInfoSolid />}
       theme="dark"
       heading="Informational alert"
       extended={true}
@@ -91,7 +91,7 @@ Warning alerts present information about the action that is about to be taken th
     <Alert
       onClose={() => undefined}
       color={colors.orange.base}
-      Icon={IconWarningSolid}
+      icon={<IconWarningSolid />}
       heading="Warning alert"
       actions={<Button size="small">Confirm</Button>}
     >
@@ -103,7 +103,7 @@ Warning alerts present information about the action that is about to be taken th
     <Alert
       onClose={() => undefined}
       color={colors.orange.base}
-      Icon={IconWarningSolid}
+      icon={<IconWarningSolid />}
       heading="Warning alert"
       theme="dark"
       actions={
@@ -127,7 +127,7 @@ Error alerts let the user know that something has gone wrong or is blocked in th
     <Alert
       onClose={() => undefined}
       color={colors.red.base}
-      Icon={IconErrorSolid}
+      icon={<IconErrorSolid />}
       heading="Error alert"
       style={{ marginBottom: 20 }}
     >
@@ -138,7 +138,7 @@ Error alerts let the user know that something has gone wrong or is blocked in th
     <Alert
       onClose={() => undefined}
       color={colors.red.base}
-      Icon={IconErrorSolid}
+      icon={<IconErrorSolid />}
       heading="Error alert"
       theme="dark"
       style={{ marginBottom: 20 }}
@@ -157,7 +157,7 @@ Success alerts are confirmation that the action the user was trying to take has
     <Alert
       onClose={() => undefined}
       color={colors.green.base}
-      Icon={IconSuccessSolid}
+      icon={<IconSuccessSolid />}
       heading="Success alert"
     >
       A new schema registry has been created by timbotnik.
@@ -167,7 +167,7 @@ Success alerts are confirmation that the action the user was trying to take has
     <Alert
       onClose={() => undefined}
       color={colors.green.base}
-      Icon={IconSuccessSolid}
+      icon={<IconSuccessSolid />}
       theme="dark"
       heading="Success alert"
     >
diff --git a/src/Alert/Alert.tsx b/src/Alert/Alert.tsx
index d22dd52..ab5c1d1 100644
--- a/src/Alert/Alert.tsx
+++ b/src/Alert/Alert.tsx
@@ -25,7 +25,7 @@ interface AlertProps {
   /**
    * The icon displayed to the left of the heading text
    */
-  Icon: React.ComponentType<{ className?: string }>;
+  icon: React.ReactElement<{ className?: string }>;
 
   heading: React.ReactNode;
 
@@ -73,7 +73,7 @@ export const Alert: React.FC<AlertProps> = ({
   headingAs = "h2",
   children,
   color,
-  Icon,
+  icon,
   theme = "light",
   extended = false,
   ...otherProps
@@ -120,9 +120,23 @@ export const Alert: React.FC<AlertProps> = ({
               ),
               children: (
                 <Fragment>
-                  <Icon
-                    css={{ width: 20, height: 20, color, marginRight: 13 }}
-                  />
+                  <ClassNames>
+                    {({ css, cx }) =>
+                      React.cloneElement(icon, {
+                        className: classnames(
+                          icon.props.className,
+                          cx(
+                            css({
+                              width: 20,
+                              height: 20,
+                              color,
+                              marginRight: 13,
+                            })
+                          )
+                        ),
+                      })
+                    }
+                  </ClassNames>
                   {heading}
                 </Fragment>
               ),
@@ -190,5 +204,5 @@ Alert.propTypes = {
       .map((color) => Object.values(color))
       .reduce((a, b) => a.concat(b)) as PaletteColor[]
   ).isRequired,
-  Icon: PropTypes.func.isRequired,
+  icon: PropTypes.element.isRequired,
 };
```
</details>

src/Alert/Alert.tsx
```diff
@@ -0,0 +1,194 @@
...
+  extended = false,
+  ...otherProps
+}) => {
+  return (
+    <section
+      {...otherProps}
+      css={{
+        backgroundColor:
+          theme === "light" ? colors.white : colors.black.lighter,
+        color: theme === "light" ? colors.black.base : colors.white,
```
justinanastos, line 87: This doesn't take into account that we might have other themes that aren't "dark" or "light", but that's not very likely so that's fine. If we wanted to protect ourselves from this then I'd make the following change:

<details>
<summary>expand diff</summary>

```diff
diff --git a/src/Alert/Alert.tsx b/src/Alert/Alert.tsx
index d22dd52..ec8e791 100644
--- a/src/Alert/Alert.tsx
+++ b/src/Alert/Alert.tsx
@@ -66,6 +66,11 @@ interface AlertProps {
   style?: CSSProperties;
 }
 
+// @see https://www.typescriptlang.org/docs/handbook/advanced-types.html#exhaustiveness-checking
+function assertUnreachable(x: never): never {
+  throw new TypeError(`Didn't expect to get here ${x}`);
+}
+
 export const Alert: React.FC<AlertProps> = ({
   heading,
   onClose,
@@ -83,8 +88,17 @@ export const Alert: React.FC<AlertProps> = ({
       {...otherProps}
       css={{
         backgroundColor:
-          theme === "light" ? colors.white : colors.black.lighter,
-        color: theme === "light" ? colors.black.base : colors.white,
+          theme === "light"
+            ? colors.white
+            : theme === "dark"
+            ? colors.black.lighter
+            : assertUnreachable(theme),
+        color:
+          theme === "light"
+            ? colors.black.base
+            : theme === "dark"
+            ? colors.white
+            : assertUnreachable(theme),
         boxShadow: `0 4px 8px 0 rgba(0, 0, 0, .04)`,
         borderStyle: "solid",
         borderRadius: 4,
```
</details>

Then if we add another option to `theme`, TypeScript will recognize we need to add cases here.

You don't have to make a change, just pointing thi sout.

src/Alert/Alert.tsx
```diff
@@ -0,0 +1,194 @@
...
+  extended: PropTypes.bool,
+  onClose: PropTypes.func.isRequired,
+  children: PropTypes.node,
+  heading: PropTypes.node.isRequired,
+  actions: PropTypes.node,
+  color: PropTypes.oneOf(
+    Object.values(colors)
+      .map((color) => Object.values(color))
+      .reduce((a, b) => a.concat(b)) as PaletteColor[]
+  ).isRequired,
```
justinanastos, line 192: Clever.

src/Alert/Alert.tsx
```diff
@@ -0,0 +1,194 @@
...
+  Icon: PropTypes.func.isRequired,
```
justinanastos, line 193: I'm asking for this to be changed elsewhere so you don't have to change this here, but this will fail the check if `Icon` is a class component.

src/Alert/Alert.tsx
```diff
@@ -0,0 +1,194 @@
...
+          {children}
+        </div>
+
+        {actions}
+      </div>
+    </section>
+  );
+};
+
+Alert.propTypes = {
```
justinanastos, line 182: This doesn't include `headingAs`, which might be intentional because it's not a fun one to type 😃 I'm fine with leaving it out.

justinanastos:
@Jephuff I also updated the title of this PR because that's what will show up in the change log 😃

justinanastos:
@cheapsteak I agree with you on coming up with something to prevent us from having to re-invent the wheel for the rounded, bordered, shadowed edge of these components. I thought it was fine to keep outside the scope of this; you're right we'll totally need to do that at some point. I hadn't actually considered that it's something we'd want outside of Space Kit. If that's the case, I'd name it something like `OutlinedBox` or something.

src/Alert/Alert.tsx
Jephuff: yep that makes a lot of sense! applied the patch 👍 
I've never actually encounter a patch before, very handy!

src/Alert/Alert.tsx
Jephuff: I'm definitely for this. 
There were a couple assertUnreachable in the codebase already so I made a common one in shared. If there was a reason they were in each file I can definitely split them up again 👍

src/Alert/Alert.tsx
Jephuff: haha, yeah it wasn't in Card so I just left it off, but might as well add it! I just did PropTypes.string because thats a long list otherwise hahah

src/Alert/Alert.story.mdx
```diff
@@ -42,7 +42,7 @@ Informational alerts present general info about something you are doing or worki
     <Alert
       onClose={() => undefined}
       color={colors.blue.base}
-      icon={<IconInfoSolid />}
+      icon={<IconInfoSolid css={{ "& .inner": { fill: colors.white } }} />}
```
justinanastos, line 45: This seems like a failing of our icon components that we should address soon.

src/Alert/Alert.tsx
```diff
@@ -0,0 +1,222 @@
...
+              className: cx(
+                css({
+                  fontWeight: 600,
+                  marginBottom: 0,
+                  marginTop: 0,
+                  width: "100%",
+                  display: "flex",
+                  color: getOffsetInPalette(
+                    2,
+                    theme === "light" ? "darker" : "lighter",
```
justinanastos, line 125: minor: Another good spot for an `assertUnreachable`.

src/Alert/Alert.tsx
```diff
@@ -0,0 +1,222 @@
...
+              theme === "dark" ? colors.grey.dark : colors.silver.dark,
+            marginTop: 14,
+            marginBottom: 14,
+          }}
+        />
+      )}
+      <div css={{ marginLeft: extended ? 0 : 33 }}>
+        <div
+          css={{
+            fontSize: 13,
```
justinanastos, line 192: minor: Use Space Kit typography for this

https://github.com/apollographql/space-kit/blob/b3fa01c71208ae0e1f0e15b32e85cf846b80a142/src/typography/base.ts#L37-L41

src/Alert/Alert.tsx
```diff
@@ -0,0 +1,222 @@
...
+          }}
+        />
+      </div>
+
+      {extended && (
+        <hr
+          css={{
+            height: 1,
+            borderWidth: 0,
+            backgroundColor:
```
justinanastos, line 182: minor: Another spot for `assertUnreachable`.

src/shared/assertUnreachable.ts
```diff
@@ -0,0 +1,5 @@
+// @see https://www.typescriptlang.org/docs/handbook/advanced-types.html#exhaustiveness-checking
+/* istanbul ignore next */
+export function assertUnreachable(value: never): never {
+  throw new TypeError(`Unreachable value reached ${value}`);
+}
```
justinanastos, line 5: I'm glad you separated this out. I was going to ask you to add this to the excluded files in rollup so we don't export an `assertUnreachable` file, but our config already handles this 😃

https://github.com/apollographql/space-kit/blob/b3fa01c71208ae0e1f0e15b32e85cf846b80a142/rollup.config.js#L32-L46

Jephuff:
@jglovier Updated the text with some more explanation 👍 
also that is a great point about the name, this will collide with alert banners. I've renamed it to AlertCard, @justinanastos  does that sound good to you?

src/AlertCard/AlertCard.story.mdx
```diff
@@ -0,0 +1,207 @@
...
+      onClose={() => undefined}
+      color={colors.green.base}
+      icon={<IconSuccessSolid css={{ "& .inner": { fill: colors.white } }} />}
+      theme="dark"
+      heading="Success alert"
+    >
+      A new schema registry has been created by timbotnik.
+    </AlertCard>
+  </Story>
+</Preview>
```
justinanastos, line 201: I missed this in my review, so I'm sorry. We're not going to share contextual components as Space Kit ideas, just components that can be configured. Instead of demonstrating like this, demonstrate how the component can be configured.

src/AlertCard/AlertCard.tsx
```diff
@@ -0,0 +1,234 @@
+/** @jsx jsx */
```
justinanastos, line 1: I'm commenting here just so the conversation can be threaded. I'm confused as to why this is named `AlertCard`. Which item in the design principals docs is this implementing and what naming conflict was there?

https://apollographql.quip.com/AgFcAwybcEWN/Alerts-design-principles

Jephuff: These are to be used in the Alert toast cards.
Joel pointed out the conflict
> The other thing I noticed is that in SpaceKit this component is listed as an Alert, but technically it's a specific type of alert (alert toast card) out of two currently designed for types (alert banner and alert toast cards). Is your intention to rename these to Alert Toast Cards or a similar designation after we add the alert banners?

We will need an AlertBanner soon so `AlertCard` and `AlertBanner` seemed preferable to `Alert` and `AlertBanner`

src/AlertCard/AlertCard.story.mdx
Jephuff: I can definitely remove the concepts of error/warning/success/error.
before I do that, @jglovier I think you might have had thoughts on this?

apollo-bot2:
:rocket: PR was released in [`v5.6.0`](https://github.com/apollographql/space-kit/releases/tag/v5.6.0) :rocket:

jglovier:
shufflin-robot

PR #179, opened 2020-06-04
AR-1689 Fix crashing chromatic builds

justinanastos:
Chromatic started crashing in CI. My instincts tell me it's a memory issue since we've seen this in the past. Use a larger container class and notify node that it can use 6GB of memory instead of 4.

Resolves [AR-1689](https://apollographql.atlassian.net/browse/AR-1689)

---

This is a stopgap solution until we update to Storybook v6 which supposedly resolves the bug causing such high memory usage https://github.com/storybookjs/storybook/issues/7743. I spent an hour unsuccessfully trying to upgrade to the latest beta. This will have to hold us over.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,7 @@ defaults: &defaults
 jobs:
   chromatic:
     <<: *defaults
+    resource_class: medium+
     steps:
       - checkout
       - run:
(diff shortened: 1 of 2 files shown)


cheapsteak:
can we disable minification instead? should fix the issue for netlify as well

justinanastos:
> can we disable minification instead? should fix the issue for netlify as well

@cheapsteak 

I don't believe the issue is minification @cheapsteak . This issue is coming from `react-docgen-typescript-loader` where it just uses incredible amounts of memory.

apollo-bot2:
:rocket: PR was released in [`v5.2.0`](https://github.com/apollographql/space-kit/releases/tag/v5.2.0) :rocket:

PR #182, opened 2020-06-05
AP-1658 Create `Popup` and `List` components

justinanastos:
Contributes to https://apollographql.atlassian.net/browse/AP-1658

## Release Notes

* Split `Menu` into `Popover` and `List` and rename `MenuHeading` to `ListHeading`, `MenuDivider` to `ListDivider`, and `MenuItem` to `ListItem`
* Rename `color` configuration in `List` (formerly in `Menu`) to `selectedColor`
* Add `hoverColor` configuration to `List`
* Add "large" `iconSize` to `List` configuration
* Add `padding` to `List` configuration
* Add `cloneElements` helper to make it easy to add `as` props
* Add `endIconAs` and `startIconAs` props to `List` and `ListItem` to allow configuring how they are rendered.
  * This is useful for adding colors and setting manual props on the elements themselves instead of nesting new elements underneath. 
  * This will now provide automatic truncation.


src/Popup/index.tsx
```diff
@@ -1,54 +1,26 @@
...
- * Menu element
- */
-export const Menu: React.FC<Props> = ({
+export const Popup: React.FC<Props> = ({
   children,
-  closeOnMenuItemClick = true,
-  color,
   fallbackPlacements,
-  iconSize,
   content,
```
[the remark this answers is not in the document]
justinanastos: I'm struggling with this @cheapsteak . I actually like this idea of using `target`, but that's already used and I can't think of another name (`target` is used internally and passed-through for tippy.js as the events to target). I'm thinking maybe I instead datamap that to `triggerEvent` and then add `trigger` as a component.

AlexanderMann:
@justinanastos where do we reference `Menu` now? It seems like most of the files pertaining to it have been removed/renamed altogether? But I'm guessing I missed where we swapped out the actual component's innards?

timbotnik:
I _think_ your references to `MenuXXX` are out of date in the PR description (they now look like `ListXXX`) - and I don't see where your question on naming fits in exactly.  All that aside, this seems sweet!  
Would it be correct to summarize this as
`<Menu><MenuItem /></Menu>` => `<Popover><List><ListItem /></List></Popover>`?

justinanastos:
> I _think_ your references to `MenuXXX` are out of date in the PR description (they now look like `ListXXX`) - and I don't see where your question on naming fits in exactly. All that aside, this seems sweet!
> Would it be correct to summarize this as
> `<Menu><MenuItem /></Menu>` => `<Popover><List><ListItem /></List></Popover>`?

That's exactly what I did @timbotnik .

Sorry @timbotnik  and @AlexanderMann ; I've been working on this and haven't updated the PR description.

src/ListConfig/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+   * @default <div />
+   */
+  endIconAs: React.ReactElement;
+
+  /**
+   * Background color of the hovered item in a menu
+   *
+   * The text color will be determined automatically
+   */
+  hoverColor: ShadedColor;
```
cheapsteak, line 28: Is this in response to https://meteor.slack.com/archives/CJM405HJB/p1592282599095400 ?  
I wonder if this might need to vary by individual items?  
E.g. some items might have a blue hover color, and some might be red, depending on what the individual item is?
pinging @Jephuff iirc you PR'ed the fix for this in the app

justinanastos:
> I think we may need to make `hoverColor` configureable per ListItem
> otherwise looks great!

You're probably right. I had to make `endIconAs` and `startIconAs` configurable via `List` and `ListItem`. I didn't abstract it in this case, but that might be something we want to follow up on with an automatic solution.

src/ListConfig/index.tsx
justinanastos: It's not; it's just with the goal of being able to make this work in dark or light mode with a single configuration option.

apollo-bot2:
:rocket: PR was released in [`v6.0.0`](https://github.com/apollographql/space-kit/releases/tag/v6.0.0) :rocket:

PR #193, opened 2020-06-19
Add xlarge icon size for list items

Jephuff:
Adds an xlarge icon size with 20px icons


Jephuff:
Going to override instead

PR #197, opened 2020-06-26
darken spinner trail by 1 shade

cheapsteak:
Pages in Studio now have grey-light backgrounds, and the orbit blends in to the background
download

Making the orbit a shade darker for the "light" theme would solve this problem

It does change the orbit trail color for an existing theme so it's a breaking change; however this seemed more pragmatic than creating a different theme name to call this

Longer term if we revist the component's api, we should consider passing the 2 colors we're using in via props (e.g. fgColor, bgColor) so we would require a version bump and package update to change colors (cui bono?)

### DIFF
--- a/src/Loaders/LoadingSpinner.tsx
+++ b/src/Loaders/LoadingSpinner.tsx
@@ -53,7 +53,7 @@ export const LoadingSpinner = React.forwardRef<SVGSVGElement, Props>(
       }
     > = {
       light: {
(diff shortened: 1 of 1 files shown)


apollo-bot2:
:rocket: PR was released in [`v7.0.0`](https://github.com/apollographql/space-kit/releases/tag/v7.0.0) :rocket:

PR #202, opened 2020-07-14
Update overflow icon

jglovier:
Tweaks styles for the overflow icon to be a little less big and puffy and a little more spacious (i.e. more standardized).

| Before | After |
|---|---|
| image |image|


jglovier:
Oddly enough, the Chromatic changes for this branch seem to be broken or not working correctly. The only diff that came up for this change was for the tooltip with normal padding (which I didn't change), and the icon view didn't actually show the diff.

Also, I'm noticing that the teal version doesn't change color, even though the fill on the path is set to `#000`. 🤔 

image

jglovier:
Nevermind, switching the fill value from `#000` to `#000000` seems to have fixed it. I thought `#000` worked too.

Jephuff:
> Nevermind, switching the fill value from `#000` to `#000000` seems to have fixed it. I thought `#000` worked too.

huh, I thought #000 would work as well, it's in the list of things to replace 🤔

jglovier:
Weird that the Chromatic tests still aren't picking up a change for the icon though. Perhaps another argument for separating icon tests out into their own stories as @justinanastos was saying.

apollo-bot2:
:rocket: PR was released in [`v7.3.0`](https://github.com/apollographql/space-kit/releases/tag/v7.3.0) :rocket:

PR #204, opened 2020-07-15
Popover: add interactive property

timbotnik:
I feel like this is more natural than stopping propagation in the descendants, but feedback welcome.  If we like this I'll go through whatever additional ceremony required to bump the version and release.


apollo-bot2:
:rocket: PR was released in [`v7.5.0`](https://github.com/apollographql/space-kit/releases/tag/v7.5.0) :rocket:

PR #206, opened 2020-07-17
Modal: add verticalScrollMode property to limit overflow to children

timbotnik:
In some cases we want to be able to see the action buttons even if the child content overflows - this gives us a way to use a flex column layout and move the `overflow-y` to the `div` that wraps `children`.

### DIFF
--- a/src/Modal/Modal.story.tsx
+++ b/src/Modal/Modal.story.tsx
@@ -213,4 +213,51 @@ storiesOf("Modal", module)
         </div>
       </Modal>
     )
-  );
+  )
+  .add("using verticalScrollMode 'modal'", () => (
+    <Modal
+      size="medium"
+      title="Modal Title"
+      primaryAction={
+        <Button color={colors.green.base} style={{ color: colors.white }}>
+          Buy 1 Seat
+        </Button>
+      }
+      secondaryAction={<Button color={colors.white}>Cancel</Button>}
+      bottomLeftText={
+        <span style={{ color: colors.blue.base }}>
+          Update Billing Information
+        </span>
+      }
+    >
+      <div>
+        {Array.from(Array(500))
+          .map(() => "lorem ipsum dolor")
+          .join(" ")}
+      </div>
+    </Modal>
+  ))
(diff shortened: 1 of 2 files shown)


timbotnik:
Thx for the insightful review @cheapsteak.  I agree with you on most of these points.

> Is there a way to solve this by adding a less specific prop? (maybe expose classNames on the div that wraps the "actions"?)

I would have liked this as well, but it seemed like I needed to do these 3 things:
1.  turn off `overflow-y: auto` on the parent div so that we'd be able to handle the overflow in descendants
2. make the layout `display: flex; flex-direction: column` so we could stretch the `children` area to fill the modal
3. set `overflow-y: auto` on the div that wraps `children`.
I think there is a slight alternative here where we could pass in `classNames` to get #1 & #2 and expose a  `childWrapperClassnames` property for #3.  It did feel pretty weird to me that we'd be passing in a flex layout which would depend on the nesting of things inside the modal to stay the same so I didn't pursue that option.  It felt like this component had already made a call to be more of a layout owner, and this could certainly lead to options creep if we keep going this way.  

I think ultimately what we might want is a way to separate a `Modal` (which handles the basic treatment of popping up a card) and the internal Modal layout as different things, allowing us to have different modal layouts or potentially fully customize the layout at usage site.

> Aside from the concern for option-bloat, verticalScrollMode="children" is a sequence of words that I might have trouble making a reasonable guess at what it does if I saw them at a usage site

I toyed around with different names here like `overflowYAt: 'children' | 'modal'`, but yeah not really the main problem here.

> I'm not sure how to balance "wanting a great api for a shared public ui lib" and "not wanting to block sprint work in the main project the lib is used for". Going to punt that question to other reviewers 🙈

Based on time we need to ship something now and revisit it.

timbotnik:
Merging but let's get this refactor on the agenda at some point.

apollo-bot2:
:rocket: PR was released in [`v7.6.0`](https://github.com/apollographql/space-kit/releases/tag/v7.6.0) :rocket:

PR #210, opened 2020-07-21
AR-1779 AR-1758 Refactor `Button` to use `react-spectrum`

justinanastos:
Resolves [AR-1779](https://apollographql.atlassian.net/browse/AP-1779) and contributes to [AR-1758](https://apollographql.atlassian.net/browse/AP-1758)

## Release Notes

* Add test for `style` in `Button`
* Add test for `className` merging in `Button`
* Use [`@react-aria/utils`'s `mergeProps`](https://react-spectrum.adobe.com/react-aria/mergeProps.html) to intelligent props merging
* Use [`@react-aria/useFocusRing`](https://react-spectrum.adobe.com/react-aria/useFocusRing.html) to handle focus state
* Use [`@react-aria/useButton`](https://react-spectrum.adobe.com/react-aria/useButton.html) to handle accessibility and the `isPressed` state
* Use [`@react-aria/interactions`](https://react-spectrum.adobe.com/react-aria/useHover.html) 's `useHover` to handle accessibility and the `isHovered` state

# Reviewing

I don't think this needs a strong review. The testing is solid and the storybook stories are robust. Click around the build preview to make sure the hover, active, and focus states work. Tab to the buttons as well to test that you get a focus state with the keyboard and not the mouse.

This should look visually identical to the previous version with the exception of keyboard navigation working correctly.

# Other Concerns

`useButton` doesn't emulate a native `<button>` in that it's intended to be an abstraction and we, as consumers, are expected to understand that. I used the same logic in `Switch` and it feels a little funny, but hopefully it'll feel like home soon enough. The biggest differences are the `disabled` prop and `onClick`; `useButton` uses `isDisabled` and `onPress` in favor of those. I set this up to accept both and to continue to accept `onClick` on the `as` prop for reverse compatibility. I deprecated those fields, but I don't really expect to ever remove them.

Resolves [AR-1779](https://apollographql.atlassian.net/browse/AR-1779)

<details>
  <summary>📦 Published PR as canary version: <code>7.16.1-canary.210.5392.0</code></summary>
  <br />
  
  ```bash
  npm install @apollo/space-kit@7.16.1-canary.210.5392.0
  # or 
  yarn add @apollo/space-kit@7.16.1-canary.210.5392.0
  ```
</details>


src/Button/Button.tsx
```diff
@@ -276,293 +279,296 @@ export const Button = React.forwardRef<HTMLElement, Props>(
...
-            }
+          const propsToPass = mergeProps(
+            // Omit `onClick` from `otherProps` beacuse we'll be conditionally
+            // calling it in the `onClick` handler depending on the `disabled`
+            // prop.
+            omit(otherProps, "onClick"),
+            mergeProps(focusProps, {
+              ref,
+              className: classnames(
+                isFocusVisible && cx(css(focusedStyles)),
```
cheapsteak, line 351: this `cx` seems redundant?

src/Button/Button.tsx
```diff
@@ -276,293 +279,296 @@ export const Button = React.forwardRef<HTMLElement, Props>(
...
+                    },
+                  ])
+                )
+              ),
+              disabled,
+              onClick: (event: React.MouseEvent<HTMLElement, MouseEvent>) => {
+                if (disabled) return event.preventDefault();
+
+                otherProps.onClick?.(event);
+                as.props.onClick?.(event);
```
cheapsteak, line 500: nbd: I wonder if we should throw when both are defined?

```tsx
<Button
  as={
    <button
      onClick={(e) => {
        e.preventDefault();
        e.stopPropagation();
      }}
    />
  }
  onClick={() => alert("expected behaviour seems unclear from reading usage site")}
/>;
```

or make it so that the click handler from the `as` component would override the one from the Button component (and warn that Button's will be ignored if both are provided?)

src/Button/Button.tsx
justinanastos: I had no idea what `cx` was doing; this is totally using it wrong everywhere! I have another commit I'm going to push that shrinks down the `<ClassNames>` insides to only relate to stuff that actually needs `ClassNames` and I'll use `cx` correctly.

src/Button/Button.tsx
justinanastos: I don't feel strongly about preventing `onClick` from going in either place. I'd honestly expect if it was in both that they would both work, but I don't feel super strongly about it.

src/Button/Button.spec.tsx
```diff
@@ -1,36 +1,33 @@
...
 import { IconShip2 } from "../icons/IconShip2";
 import React from "react";
+import userEvent from "@testing-library/user-event";
 
 test("default button renders", () => {
-  const { getByText } = render(<Button>submit</Button>);
+  render(<Button>submit</Button>);
 
-  expect(getByText("submit")).toBeInTheDocument();
+  expect(screen.getByRole("button")).toBeInTheDocument();
```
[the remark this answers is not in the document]
justinanastos: Yes we should, even though there's no problem _not_ doing it. That's just how they recommend it and it's more ergonomic in my opinion.

package.json
```diff
@@ -58,15 +58,21 @@
...
     "@react-aria/visually-hidden": "^3.1.0",
     "@react-stately/toggle": "^3.1.0",
     "@tippyjs/react": "^4.0.0",
     "@types/classnames": "^2.2.9",
+    "@types/lodash": "^4.14.157",
     "@types/tinycolor2": "^1.4.2",
     "classnames": "^2.2.6",
+    "i": "^0.3.6",
+    "lodash": "*",
+    "npm": "^6.14.7",
```
cheapsteak, line 75: 🧹

rollup.config.js
```diff
@@ -60,6 +64,7 @@ function CJS() {
       "classnames",
       "framer-motion",
       "prop-types",
+      "lodash/omit",
```
cheapsteak, line 67: 🧹

package.json
justinanastos: 🤦

justinanastos:
> Any idea why some but not _all_ of the stories show a focus ring on the button when tabbing?

@timbotnik which ones don't show a focus ring? I believe all the buttons show a focus ring except when they're disabled.

justinanastos:
Closing this because we're not going to follow this approach.

PR #211, opened 2020-07-23
AR-1781 Don't render `Tooltip`, `ConfirmationTooltip`, or `Popover` if `content` prop is `undefined`

justinanastos:
Resolves issue [AR-1781](https://apollographql.atlassian.net/browse/AP-1781)

### DIFF
--- a/src/ConfirmationTooltip/ConfirmationTooltip.story.tsx
+++ b/src/ConfirmationTooltip/ConfirmationTooltip.story.tsx
@@ -28,6 +28,27 @@ storiesOf("ConfirmationTooltip", module)
       chromatic: { delay: 500 },
     }
   )
+  .add(
+    "disabled, no confirmation should be visible",
+    () => {
+      return (
+        <PerformUserInteraction
+          callback={async () => {
+            userEvent.click(await findByText(document.body, "click"));
+          }}
+        >
+          <div style={{ padding: 50 }}>
+            <ConfirmationTooltip content="content" disabled>
+              <Button>click</Button>
+            </ConfirmationTooltip>
+          </div>
+        </PerformUserInteraction>
+      );
+    },
+    {
+      chromatic: { delay: 500 },
+    }
+  )
   .add("interactive", () => {
     return (
       <div style={{ padding: 50 }}>
--- a/src/ConfirmationTooltip/index.tsx
+++ b/src/ConfirmationTooltip/index.tsx
@@ -4,7 +4,7 @@ import { AbstractTooltip, AbstractTooltipProps } from "../AbstractTooltip";
 
 type Props = Pick<
   React.ComponentProps<typeof AbstractTooltip>,
-  "children" | "content"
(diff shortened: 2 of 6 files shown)


Jephuff:
Nice! I've needed this a couple times :)

Could we expose the `disabled` prop on AbstractTooltip from tippy as an alternative? I could see some confusion if something is undefined and clicking the menu just does nothing, might take a little longer to track down, but an empty menu tells you exactly where to go.

src/AbstractTooltip/index.tsx
```diff
@@ -49,6 +49,10 @@ export const AbstractTooltip: React.FC<Props> = ({
   const disableAnimations =
     disableAnimationsFromProvider || forceVisibleForTestingOnly;
 
+  if (!props.content) {
+    return children ?? null;
```
jgzuke, line 53: Can we switch this to also check for empty array? Right now I think this would still show up for
```jsx
<>
  {false && ...}
  {false && ...}
</>
```

justinanastos:
> Nice! I've needed this a couple times :)
> 
> Could we expose the `disabled` prop on AbstractTooltip from tippy as an alternative? I could see some confusion if something is undefined and clicking the menu just does nothing, might take a little longer to track down, but an empty menu tells you exactly where to go.

I think this is what @jgzuke originally asked for, would that be better for all these use cases?

src/AbstractTooltip/index.tsx
jgzuke: Can we also change this on `Card`? I think it our graph integrations page this is causing a bunch of extra margins

justinanastos:
Then a `disabled` alternative makes way more sense :-D I was not in favor of it, thanks for bearing with me @jgzuke and @Jephuff

justinanastos:
Updated to use `disabled` prop which is apparently already supported by the underlying `tippy.js` library.

src/AbstractTooltip/index.tsx
justinanastos: I'm going to stick with using the `disabled` prop and I feel that an empty card should continue to render itself. In this case, this isn't an empty array, it's a Fragment that it's own children.

This is a different case because rendering content with or without the tooltip is not an easy change, as you know because you requested this 😃

jgzuke: Ah sorry this was a different thing. We add padding to the bottom of the card only if the card has children, I think we use this same style check. We should fix it to check for empty arrays as well

justinanastos: > Ah sorry this was a different thing. We add padding to the bottom of the card only if the card has children, I think we use this same style check. We should fix it to check for empty arrays as well

Ah I better understand now. That makes sense.

apollo-bot2:
:rocket: PR was released in [`v7.7.0`](https://github.com/apollographql/space-kit/releases/tag/v7.7.0) :rocket:

PR #212, opened 2020-07-24
AR-748 Add `Switch` component

justinanastos:
Please check the storybook deploy preview and visit Switch's docs tab for usage examples!

This is intended to provide a minimum viable abstraction for a Switch component. All use cases are documented (hopefully) in the storybook docs. The props are taken directly from `react-spectrum`, which feels fine to me. We also make available `Toggle.Label` if you want to use the `label` separately just for rendering something that looks like a toggle label. When using this, you should wire up `aria-labeledby` on your own.

This was originally titled `Toggle`, but the underlying `<input>` 's role should be `role="switch"`, so I decided to name this `Switch` instead. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role

Contributes to [AR-748](https://apollographql.atlassian.net/browse/AP-748)

Depended on by mdg-private/studio-ui#2953


src/Checkbox/Checkbox.story.mdx
```diff
@@ -0,0 +1,155 @@
...
+  of={CheckboxComponent}
+  markdown="**Checkboxs** use the same props schema as the underlying library from react-spectrum."
+/>
+
+## Uncontrolled
+
+Using the `defaultSelected` props to indicate these components are uncontrolled; meaning their state is not stored externally.
+
+<Preview>
+  <Story name="Uncontrolled Deselected">
```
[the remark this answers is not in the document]
justinanastos: Click on the `Docs` tab up top to see them all in one page with longform documentation @cheapsteak 

image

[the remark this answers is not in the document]
cheapsteak: Ahhhh missed that, thanks!

justinanastos:
@cheapsteak I'm terrible at framer motion. I couldn't get the animation to look smooth with a spring; I really miss react-spring's presets that let me just choose "stiff" instead of "bouncy". Any tips there would be appreciated.

src/SpaceKitProvider/index.tsx
```diff
@@ -9,10 +9,13 @@ interface State {
    * @default false
    */
   disableAnimations: boolean;
+
+  theme: "light" | "midnight";
```
Jephuff, line 13: Can we use `dark` instead of `midnight` to keep it consistent with our theme names everywhere else?

justinanastos: Thanks for starting this conversation. After talking with @jglovier we should rename this to `dark` because using midnight is an implementation detail 👍

justinanastos: Thanks for starting this conversation @Jephuff ! After talking with @jglovier we should rename this to `dark` because using midnight is an implementation detail 👍

justinanastos: Done.

src/Switch/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+              position: "relative",
+              width: (8 / 3) * dotSize,
+            })}
+          >
+            <motion.div
+              animate={{
+                x: state.isSelected ? dotSize + borderSize * 2 : 0,
+              }}
+              initial={false}
+              transition={{ duration: 0.5, type: "tween", ease: "easeOut" }}
```
jglovier, line 175: Can we speed this up a little, say 150ms?
```suggestion
              transition={{ duration: 0.15, type: "tween", ease: "easeOut" }}
```

jglovier:
Looking great!! :zap: :zap: :zap:

I noticed that there is a disabled state in the Story docs, but it's a little different than I'd expect the disabled state to look. I would assume that disabled would mean the setting is off and cannot be turned on. Also, the disabled state could use a lighter opacity on the labels as well, so it's clear that the setting is disabled and not just the toggle switch.

justinanastos:
Hey @jglovier !

> Looking great!! ⚡ ⚡ ⚡
> 
> I noticed that there is a disabled state in the Story docs, but it's a little different than I'd expect the disabled state to look. I would assume that disabled would mean the setting is off and cannot be turned on. Also, the disabled state could use a lighter opacity on the labels as well, so it's clear that the setting is disabled and not just the toggle switch.

The dark-mode disabled state is definitely not what it should be because I don't have any examples of what exactly it should be 😃 I imagined the disabled state to just mean you can't change it, not that it was selected or unselected. I can mess with the opacity and we can come back to it if it lands like this 😃

justinanastos:
@jglovier I'm going to change the animation to copy the animation in Modal because that seems to be configured to be pretty slick. It feels better and is much faster now.

Regarding the disabled color, I def need some feedback on it; namely what should it look like in dark mode. Just changing the opacity looks wrong because then it's blending with the dark background color. Right now the color of the selected state gets lighter when it _should_ get darker; but I don't think this is a blocker for merging and releasing since we're not using it yet and I'm anxious :-D If it's something that we can put together quickly then I'm happy to do it now!

jglovier:
>I imagined the disabled state to just mean you can't change it, not that it was selected or unselected.

Ohhh, I guess I misunderstood the intent. What would be an example of where the toggle would appear, but the user would not be able to change it?

justinanastos:
> > I imagined the disabled state to just mean you can't change it, not that it was selected or unselected.
> 
> Ohhh, I guess I misunderstood the intent. What would be an example of where the toggle would appear, but the user would not be able to change it?

@jglovier I can't think of any!

justinanastos:
> > I imagined the disabled state to just mean you can't change it, not that it was selected or unselected.
> 
> Ohhh, I guess I misunderstood the intent. What would be an example of where the toggle would appear, but the user would not be able to change it?

@jglovier I thought of one. A form that has several switches and you want to disable them all while a mutation is in-flight.

jglovier:
> > > I imagined the disabled state to just mean you can't change it, not that it was selected or unselected.
> > 
> > 
> > Ohhh, I guess I misunderstood the intent. What would be an example of where the toggle would appear, but the user would not be able to change it?
> 
> @jglovier I can't think of any!

Maybe we don't need a disabled state? If the user cannot change it, I would suggest we always simply don't show it if the user doesn't have access to modify.

If that is not preferable, however, here is my suggestion for tweaking...

>Regarding the disabled color, I def need some feedback on it; namely what should it look like in dark mode. Just changing the opacity looks wrong because then it's blending with the dark background color.

Changing opacity to something like 50% for the entire element (i.e. setting name, state label, and toggle switch) is preferable. It's a relatively common pattern to use when something is disabled, and it's what we're kind of trying to simulate by changing colors but instead of simulating we're actually doing it! 😄 I would just also add a tooltip to say "You cannot change this setting."

Tailwinds appears to support `opacity-50`, although apparently we don't have it in our Tailwinds.

justinanastos:
:rocket: PR was released in [`v7.8.0`](https://github.com/apollographql/space-kit/releases/tag/v7.8.0) :rocket:

PR #214, opened 2020-08-04
Upgrade storybook to v6

justinanastos:
This is the latest and greatest RC with built-in TypeScript support.

We made this change because we had to upgrade to v5.3 to be able to override the `Preview` component, which will be step 1 to passing the theme down via context to descendent components later. While I upgraded to v5.3 I ran into issues with TypeScript docgen that was resolved with ugprading to v6, so I just went all the way.


justinanastos:
:rocket: PR was released in [`v7.8.0`](https://github.com/apollographql/space-kit/releases/tag/v7.8.0) :rocket:

PR #217, opened 2020-08-14
AR-1818 Modal: Make `as` prop propogate `ref`

justinanastos:
Resolves issue [AR-1818](https://apollographql.atlassian.net/browse/AP-1818)

Modals didn't take `as` prop's `ref`, now it does. We had to strip some type safety that we weren't actually using, which is fine.

Also adds a test to make sure this works 😃

### DIFF
--- a/src/Modal/Modal.tsx
+++ b/src/Modal/Modal.tsx
@@ -1,8 +1,7 @@
-/* eslint-disable @typescript-eslint/ban-ts-ignore */
 /** @jsx jsx */
 import { jsx, css, ClassNames } from "@emotion/core";
 import React, { useEffect } from "react";
-import { motion } from "framer-motion";
+import { motion, MotionProps } from "framer-motion";
 import * as typography from "../typography";
 import { colors } from "../colors";
 import * as CSS from "csstype";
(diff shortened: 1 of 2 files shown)


src/Modal/Modal.tsx
```diff
@@ -199,18 +203,11 @@ export const Modal: React.FC<Props> = ({
                 velocity: 8,
               },
             }}
-            onClick={
-              // Ignore the type of `event` because the React.MouseEvent
-              // generic will be a union of 170 different DOM elements;
-              // TypeScript can't handle that.
-              //
-              // @ts-ignore
```
jgzuke, line 207: This works fine now?

src/Modal/Modal.tsx
```diff
@@ -172,22 +172,26 @@ export const Modal: React.FC<Props> = ({
...
   /**
    * Framer motion component to render. The type will be taken from the `as`
-   * prop
+   * prop.
+   *
+   * We have to strip the type because we'll get an error about the union being
+   * too complex to model because `motion` has over 100 options.
    */
-  const MotionComponent = motion[type];
+  const MotionComponent = motion[type] as React.ComponentType<any>;
```
Jephuff, line 180: NDB: would unknown work here? (and typing the variable instead of the assertion?)
```suggestion
  const MotionComponent: React.ComponentType<unknown> = motion[type];
```

justinanastos: Thinking about this a little more I looked into the union type that I removed (`motion`). They implement `MotionProps` that I can use here and then the motion props will be typed correctly!

justinanastos: Also, 👍 for typing the variable instead of the assertion.

src/Modal/Modal.tsx
justinanastos: Yes. We're widening `MotionComponent`'s type to be `React.ComponentType<MotionProps>` instead of the union of all motion components so that we can get the typing we need without overloading TypeScript with a union of 170 DOM elements.

src/Modal/Modal.tsx
Jephuff: oohh thats even better! :)

apollo-bot2:
:rocket: PR was released in [`v7.10.0`](https://github.com/apollographql/space-kit/releases/tag/v7.10.0) :rocket:

PR #218, opened 2020-08-19
Fix Card bottom margin with falsey children

jgzuke:
When a card has multiple children that end up falsey, or an empty list of children this would previously add a padding to the bottom of the card. We should only add the padding if there are non falsey children

### DIFF
--- a/src/Card/Card.tsx
+++ b/src/Card/Card.tsx
@@ -61,7 +61,17 @@ export const Card: React.FC<CardProps> = ({
       paddingBottom: size === "large" ? 16 : 28,
     }}
   >
-    <div css={{ display: "flex", marginBottom: children ? 24 : 0 }}>
+    <div
+      css={{
+        display: "flex",
+        marginBottom:
+          !children ||
+          (Array.isArray(children) &&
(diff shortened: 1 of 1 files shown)


src/Card/Card.tsx
```diff
@@ -61,7 +61,17 @@ export const Card: React.FC<CardProps> = ({
...
-    <div css={{ display: "flex", marginBottom: children ? 24 : 0 }}>
+    <div
+      css={{
+        display: "flex",
+        marginBottom:
+          !children ||
+          (Array.isArray(children) &&
+            children.filter((child) => !!child).length === 0)
+            ? 0
+            : 24,
```
Jephuff, line 72: Could simplify with React.Children.toArray?

```suggestion
        marginBottom: React.Children.toArray(children).some(Boolean) ? 24 : 0
```

jgzuke: Much better! Dont know how I ended up with this 😛

apollo-bot2:
:rocket: PR was released in [`v7.9.0`](https://github.com/apollographql/space-kit/releases/tag/v7.9.0) :rocket:

PR #220, opened 2020-09-01
Add relative to label so VisuallyHidden is inside instead of under it

Jephuff:
Browser tries to scroll to input, but VisuallyHidden was way off screen. Thing makes the container relative so that the input is positioned inside. 

Fixes this UI bug
image


apollo-bot2:
:rocket: PR was released in [`v7.12.0`](https://github.com/apollographql/space-kit/releases/tag/v7.12.0) :rocket:

PR #257, opened 2020-11-06
Button: Add 4 more pixels of padding to the right of icons

justinanastos:
<details>
  <summary>📦 Published PR as canary version: <code>8.2.1-canary.257.6522.0</code></summary>
  <br />
  
  ```bash
  npm install @apollo/space-kit@8.2.1-canary.257.6522.0
  # or 
  yarn add @apollo/space-kit@8.2.1-canary.257.6522.0
  ```
</details>

### DIFF
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -519,7 +519,7 @@ export const Button = React.forwardRef<HTMLElement, Props>(
                   {icon && (
                     <ButtonIcon
                       iconSize={iconSize}
-                      className={css({ margin: iconOnly ? 0 : "0 4px 0" })}
+                      className={css({ margin: iconOnly ? 0 : "0 8px 0 0" })}
                     >
                       {icon}
                     </ButtonIcon>


src/Button/Button.tsx
```diff
@@ -519,7 +519,7 @@ export const Button = React.forwardRef<HTMLElement, Props>(
                   {icon && (
                     <ButtonIcon
                       iconSize={iconSize}
-                      className={css({ margin: iconOnly ? 0 : "0 4px 0" })}
+                      className={css({ margin: iconOnly ? 0 : "0 8px 0 4px" })}
```
Jephuff, line 522: While we are in here, do we need the `4px` on the left?

Jephuff (approved):

src/Button/Button.tsx
jglovier: Thanks for pointing this out, @Jephuff. I just noticed that there is a bit of a discrepancy between the icon spacing in the Space Kit design file:

image

...and in our current implementation in Storybook:

image

What makes this kind of difficult to both notice and address is that in the design file the button uses two words, but in the storybook version it uses just a single short word. So the question is what is the correct spacing?

Visually, I think it looks cleanest to have equal spacing between the icon and the left edge and the icon and the start of the word (or just a hair more between the icon and the button edge, if any).

But I also see that the buttons have a `min-width` applied to them, which makes this a bit tricker. Is there any reasons we need the min width on these?

Regardless of the answer to that, I'm in favor of @Jephuff's suggestion to drop the `4px` on the left. 👍

jglovier: PS - case in point where this came up as awkward was in [this thread for updating the new graph button](https://github.com/mdg-private/studio-ui/pull/3400#discussion_r519037278), and the spacing looks weird:

image

justinanastos: Thanks for this input! I'll axe that 4px and then send another PR testing getting rid of the min-width!

apollo-bot2:
:rocket: PR was released in [`v8.3.0`](https://github.com/apollographql/space-kit/releases/tag/v8.3.0) :rocket:

PR #270, opened 2020-11-10
Add support for React 17

justinanastos:
## Release Notes

React 17's release notes explaining there is a [new jsx transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html). This causes issues with emotion using it's own custom jsx transform to make the css property work. I was able to get the babel plugin that uses `css` across the board; but instead opted to make the small change of just opting out of using the new jsx transform. While this seems super easy, it wasn't, because we were using an out-of-the-box setup for both storybook and rollup, both of which had to be customized to opt out of the new jsx transform. 

In the end, this should be a transparent change.

<details>
  <summary>📦 Published PR as canary version: <code>8.1.1-canary.270.6405.0</code></summary>
  <br />
  
  ```bash
  npm install @apollo/space-kit@8.1.1-canary.270.6405.0
  # or 
  yarn add @apollo/space-kit@8.1.1-canary.270.6405.0
  ```
</details>

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -12,4 +12,25 @@ module.exports = {
     ),
     "@storybook/addon-docs",
   ],
+  babel(babelConfig) {
+    return {
+      ...babelConfig,
+      presets: babelConfig.presets.map((preset) => {
+        const presetName = Array.isArray(preset) ? preset[0] : preset;
+        const presetOptions = Array.isArray(preset) ? preset[1] : [];
+
+        if (presetName !== require.resolve("@babel/preset-react")) {
+          return preset;
+        }
+
+        return [
+          presetName,
+          {
+            ...presetOptions,
+            runtime: "classic",
+          },
+        ];
+      }),
+    };
+  },
 };
--- a/package.json
+++ b/package.json
@@ -139,6 +139,8 @@
     "npm-run-all": "^4.1.5",
     "npm-watch": "^0.6.0",
     "prettier": "^2.1.1",
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1",
     "rimraf": "^2.6.3",
     "rollup": "^1.27.8",
     "rollup-plugin-multi-input": "^1.0.3",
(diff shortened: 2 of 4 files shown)


.storybook/main.js
```diff
@@ -12,4 +12,25 @@ module.exports = {
...
+  babel(babelConfig) {
+    return {
+      ...babelConfig,
+      presets: babelConfig.presets.map((preset) => {
+        const presetName = Array.isArray(preset) ? preset[0] : preset;
+        const presetOptions = Array.isArray(preset) ? preset[1] : [];
+
+        if (presetName !== require.resolve("@babel/preset-react")) {
+          return preset;
+        }
```
Jephuff, line 24: NBD
Could return immediately, makes it a little more obvious this is only for preset-react
```suggestion
        if (presetName !== require.resolve("@babel/preset-react")) {
          return preset;
        }

        const presetName = Array.isArray(preset) ? preset[0] : preset;
        const presetOptions = Array.isArray(preset) ? preset[1] : [];
```

package.json
```diff
@@ -149,11 +155,11 @@
     "ts-jest": "^26.4.4",
     "ts-loader": "^8.0.11",
     "ts-node": "^9.0.0",
-    "typescript": "^4.0.5"
+    "typescript": "^4.1.1-rc"
```
Jephuff, line 158: Is this required for the react 17 upgrade?

Jephuff (approved):

package.json
justinanastos: No; I installed this trying to figure out why everything broke 😃 Then I decided to leave it 🤷‍♂️

.storybook/main.js
justinanastos: This will use `presetName` before it's defined; what's the advantage of this?

Jephuff: 🤦 misread that haha, disregard!

apollo-bot2:
:rocket: PR was released in [`v8.2.0`](https://github.com/apollographql/space-kit/releases/tag/v8.2.0) :rocket:

PR #274, opened 2020-11-16
Button: Remove min-width

justinanastos:
@jglovier ; maybe we need buttons with longer names than "Rest"? These just look funny.

<details>
  <summary>📦 Published PR as canary version: <code>8.2.1-canary.274.6538.0</code></summary>
  <br />
  
  ```bash
  npm install @apollo/space-kit@8.2.1-canary.274.6538.0
  # or 
  yarn add @apollo/space-kit@8.2.1-canary.274.6538.0
  ```
</details>

### DIFF
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -519,7 +519,7 @@ export const Button = React.forwardRef<HTMLElement, Props>(
                   {icon && (
                     <ButtonIcon
                       iconSize={iconSize}
-                      className={css({ margin: iconOnly ? 0 : "0 4px 0" })}
+                      className={css({ margin: iconOnly ? 0 : "0 8px 0 0" })}
                     >
                       {icon}
                     </ButtonIcon>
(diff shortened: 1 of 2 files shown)


jglovier:
So I can see how the min-width is helpful for the SpaceKit docs. Is it possible to just have the min-width set on the story versions, but not included in the component itself that ends up in Studio?

jglovier:
Also, maybe we just need a little more x padding on the buttons. 🤔

jglovier:
> Also, maybe we just need a little more x padding on the buttons. 🤔

Yeah, I think that's what is odd. They have more Y padding than X padding (or the visual equivalent), which feels very off. Most buttons have more spacing on the horizontal sides than on the vertical sides. Adding a little more padding to the sides would help.

And also, it'd be nice to be able to use min-width when necessary. (maybe as an optional itemprop?!)

jglovier (commented):
Okay GitHub Slackbot is going to keep telling me I need to respond to this unless I reply with an actual review, so that's what this is. 😛

PR #287, opened 2020-12-10
Create `FormControl` component

justinanastos:
Create a new modular `FormControl` component. 

## Reviewing

I don't really need a review here. I'm super open to any and all feedback; but it's more important to me to get this in people's hands ASAP. I'm going to merge this pretty quickly to help people start using it.

### `FormControl`

There is a new component to handle forms; `FormControl`. It's a layout component that will automatically layout it's children.

New components include: `FormLabel`, `FormDescription`, `FormInput`, `FormHelperText`, `FormStartAdornment`, `ForEndAdornment`, and `FormErrorMessage`. All of these components can be rendered anywhere on their own; but if they are rendered inside of a `FormControl` then they will automatically communicate with `FormControl` to add accessibility props and lay them out correctly.

There is a new hook, `useFormControlContext` intended to be used by any user-component that needs access to information passed by `FormControl`. `Input` and `Select` both consume `useFormControlContext`, for example.

### `Select`

  * Backtrack and rename `initialValue` prop to `defaultValue` to be inline with React's built-ins
  * Add `listAs` prop
  * Use `FormControlContext` to accept `error` state

### `Input`

New component to apply custom styles to `<input />`.

<details>
  <summary>📦 Published PR as canary version: <code>8.7.3-canary.287.7275.0</code></summary>
  <br />
  
  ```bash
  npm install @apollo/space-kit@8.7.3-canary.287.7275.0
  # or 
  yarn add @apollo/space-kit@8.7.3-canary.287.7275.0
  ```
</details>


src/FormHelperText/index.tsx
```diff
@@ -0,0 +1,69 @@
+/** @jsx jsx */
+/** @jsxFrag React.Fragment */
+import * as React from "react";
+import * as typography from "../typography";
+import { colors } from "../colors";
+import { css, jsx } from "@emotion/core";
+import { IconInfoSolid } from "../icons/IconInfoSolid";
```
jglovier, line 7: Can we use `IconInfo` here instead pls? 🙏 
```suggestion
import { IconInfo } from "../icons/IconInfo";
```

src/FormHelperText/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+      <div
+        className={className}
+        css={css({
+          ...typography.base.small,
+          color: colors.grey.base,
+          display: "flex",
+        })}
+      >
+        {showIcon ? (
+          <IconInfoSolid
```
jglovier, line 43: ...and here:
```suggestion
          <IconInfo
```

jglovier (commented):
Awesome!! Just one one request about the info icon...

src/FormHelperText/index.tsx
justinanastos: Sure can @jglovier !

jglovier (approved):
Gah, I'm sorry, I forgot to specify they should be `text-grey` instead of `text-blue`. 🙈 

Approving anyway so feel free to merge with that tweak. :bow:

src/FormControl/index.tsx
```diff
@@ -0,0 +1,262 @@
...
+                        position: "absolute",
+                        display: "inline-flex",
+                        left: 12,
+                        top: "50%",
+                        transform: "translateY(-50%)",
+                      }}
+                    >
+                      {startAdornment}
+                    </div>
+                  )}
```
cheapsteak, line 204: (Please do merge at will / don't let this slow down delivery)    

The styles around these wrappers for adornments don't look like they can be easily/safely overridden  

Could FormStartAdornment and FormEndAdornment export styled divs that carried these styles by default?

So instead of directly returning children, they could return these styled div that wraps children

src/FormStartAdornment/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+}: {
+  children: React.ReactNode;
+}): React.ReactNode {
+  const { setStartAdornment } = useFormControlInternalContext();
+
+  React.useLayoutEffect(() => {
+    setStartAdornment?.(children);
+  }, [children, setStartAdornment]);
+
+  return setStartAdornment ? null : children;
```
cheapsteak, line 21: (fleshing out what I meant above)
```suggestion
  return (
    <div
      css={{
        position: "absolute",
        display: "inline-flex",
        left: 12,
        top: "50%",
        transform: "translateY(-50%)",
      }}
    >
      {children}
    </div>
  );
```

src/FormControl/index.tsx
justinanastos: Yeah you're right that the adornment layout can't be overridden here. That's kind of unintentional and intentional at the same time. I wanted the components themselves to not have any layout styles in them; just what's needed to render them in isolation and then `FormControl` would handle the layout. I provided escape hatches for some of the layout, but obviously not this part.

I think this will need to be re-worked, especially where we don't even make an attempt to measure the size of the adornments to pad the `Input` and they just overlap the `Select`. 

I want to come back to this when we have a use case for it if that's cool.

jglovier:
break-it-down

src/FormControl/index.tsx
cheapsteak: Yup of course, merge at will  

If you don't mind, I am curious to fill out my mental model so I can better know what kind of feedback is useful to provide -- is the motivation for punting on making this portion of the component open for extension due to a desire to defer coding time? My mental model might be missing some pieces to better weigh the tradeoffs of implementing this now

cheapsteak: From my perspective of consuming components, since iterating on the API for external dependencies is not quite frictionless, if/when a use case for this arises, I would prefer if the convenient thing to do also happened to be the right thing to do. 

If these adornment components' styles are not open to extension/override, then the convenient thing becomes to wrap the child with negative margins to work around that, which is more convenient from a time-and-effort perspective, but the resulting code would feel less than great

justinanastos: @cheapsteak I was thinking in terms of "what's the use case and how should we prepare for that?" Going with a straightforward approach of just applying these css properties in `FormStartAdornment` and `FormEndAdornment` and allowing a `className` override seems fair and frictionless so I added it; thanks for the pushback!

src/FormStartAdornment/index.tsx
justinanastos: Done!

apollo-bot2:
:rocket: PR was released in [`v8.8.0`](https://github.com/apollographql/space-kit/releases/tag/v8.8.0) :rocket:

PR #298, opened 2021-01-21
AR-2016 Allow `Tooltip` to wrap a `Button` that is `disabled`

justinanastos:
This works by capturing pointer events to signal that the cursor is over the button and then applying `disabled` styles directly to the `button` element and removing the `disabled` property so that tippy.js will show the tooltip. I went with this approach because then we can still test for `disabled` buttons.

AR-2016

Testing to see how this looks in Studio UI https://github.com/mdg-private/studio-ui/pull/3700

<details>
  <summary>📦 Published PR as canary version: <code>8.12.1-canary.298.7907.0</code></summary>
  <br />
  
  ```bash
  npm install @apollo/space-kit@8.12.1-canary.298.7907.0
  # or 
  yarn add @apollo/space-kit@8.12.1-canary.298.7907.0
  ```
</details>

### DIFF
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -1,10 +1,12 @@
+/** @jsx jsx */
+/** @jsxFrag React.Fragment */
 import { colors, ShadedColor } from "../colors";
 import type { Property, SimplePseudos } from "csstype";
 import { base } from "../typography";
-import { ClassNames, ObjectInterpolation } from "@emotion/core";
+import { ClassNames, jsx, ObjectInterpolation } from "@emotion/core";
 import { getOffsetInPalette } from "../colors/utils/getOffsetInPalette";
 import tinycolor from "tinycolor2";
-import React from "react";
+import React, { useState } from "react";
 import { LoadingSpinner } from "../Loaders";
 import { assertUnreachable } from "../shared/assertUnreachable";
 import { useFocusRing } from "@react-aria/focus";
@@ -12,7 +14,7 @@ import { mergeProps } from "@react-aria/utils";
 import omit from "lodash/omit";
 import { ButtonIcon } from "./button/ButtonIcon";
 import { inputHeightDictionary } from "../shared/inputHeightDictionary";
-
+import { useTooltipContext } from "../shared/TooltipContext";
 type TLength = string | 0 | number;
 
 /**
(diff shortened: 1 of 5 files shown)


src/Button/Button.tsx
```diff
@@ -361,37 +403,34 @@ export const Button = React.forwardRef<HTMLElement, Props>(
...
             // our own with `cx`. This is necessary because `cx` allows for
             // emotion styles to be logically overwritten.
-            omit(mergedProps, "className", "onClick"),
+            omit(
+              mergedProps,
+              "className",
+              "onClick",
+              // If we're overriding the default disabled behavior, then strip
+              // it out from the props we'll pass to the element.
+              overrideDisabledBehavior ? "disabled" : "",
```
Jephuff, line 416: Is there an aria prop we should/could pass to make sure the button is still semantically disabled?

src/Tooltip/index.tsx
```diff
@@ -21,14 +22,19 @@ type Props = Pick<
...
-    </AbstractTooltip>
+    <TooltipContextProvider descendsFromTooltip>
+      <AbstractTooltip
+        delay={150}
+        trigger="mouseenter"
+        plugins={[mouseRest]}
+        reference={ref.current}
+        {...props}
+      >
+        <span ref={ref}>{children}</span>
```
Jephuff, line 36: I'm a little worried wrapping this is a span may break some of studio. We need this to get the mouse over events correct? 
Maybe we could do an alpha and test it out in studio before releasing :)

justinanastos: This is totally a mistake that I forgot to delete. Thanks for finding this!

src/Button/Button.tsx
justinanastos: Only kind-of. We can use `aria-disabled`, but this will only provide special behavior for screen readers. I'll add it.

Jephuff: ooh nice, thats probably where it matters most so thats perfect :)

Jephuff (approved):

apollo-bot2:
:rocket: PR was released in [`v8.13.0`](https://github.com/apollographql/space-kit/releases/tag/v8.13.0) :rocket:

PR #300, opened 2021-01-22
AP-1948 Add `renderListItem` prop to `Select`

justinanastos:
This will allow us to customize how the `ListItem` is rendered by `Select`,
allowing us to use something like `react-router`'s `Link`

AP-1948

Tested in Studio UI https://github.com/mdg-private/studio-ui/pull/3672

<details>
  <summary>📦 Published PR as canary version: <code>8.11.2-canary.300.7827.0</code></summary>
  <br />
  
  ```bash
  npm install @apollo/space-kit@8.11.2-canary.300.7827.0
  # or 
  yarn add @apollo/space-kit@8.11.2-canary.300.7827.0
  ```
</details>

### DIFF
--- a/src/Select/index.tsx
+++ b/src/Select/index.tsx
@@ -28,7 +28,49 @@ export type OptionProps = React.DetailedHTMLProps<
   React.OptionHTMLAttributes<HTMLOptionElement>,
   HTMLOptionElement
 >;
-interface ListItemWrapperProps extends Pick<Props, "selectionIndicator"> {
+
+// This is defined in it's own interface so that both interfaces that will use
+// it can `extend` this interface, therefore automatically including the jsdoc
+// in both types.
+interface RenderListItemProps {
+  /**
+   * Custom function to render each `ListItem`
+   *
+   * This is provided so you can render the `ListItem` on your own, with an `as`
+   * prop, for example, so you can render a `Link`. 
+   *
+   * The function will be called and it's return value rendered; this does not
+   * use `React.createElement`, so an inline function is totally acceptable with
+   * no performance penalty.
+   *
+   * @default `(props) => <ListItem {...props} />`
+   *
+   *
+   * @param props - Props that were going to be passed to the underlying
+   *`ListItem`. You must merge this with whatever you are going to render.
+
+   * @param optionElement - The `option` element that is going to be parsed and
+   * rendered as a `ListItem`.
+   *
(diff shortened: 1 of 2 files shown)


src/Select/index.tsx
```diff
@@ -64,21 +107,26 @@ const ListItemWrapper: React.FC<ListItemWrapperProps> = ({
...
-      }
-    >
-      {element.props.children}
-    </ListItem>
+    <ClassNames>
+      {({ css }) => {
+        return renderListItem(
+          {
+            className: css({ alignItems: "baseline" }),
+            key: element.props.value || element.props.children,
```
cheapsteak, line 115: NBD - is it safe to use value as the key? (Are they guaranteed unique?)

src/Select/index.tsx
```diff
@@ -64,21 +107,26 @@ const ListItemWrapper: React.FC<ListItemWrapperProps> = ({
...
-        ) : undefined
-      }
-    >
-      {element.props.children}
-    </ListItem>
+    <ClassNames>
+      {({ css }) => {
+        return renderListItem(
+          {
+            className: css({ alignItems: "baseline" }),
```
cheapsteak, line 114: NBD - am I parsing correctly that classnames on the ListItem that's passed in will be replaced rather than merged?

cheapsteak (approved):
👍

src/Select/index.tsx
cheapsteak: Oops nope I think I parsed wrong, the ListItem isn't passed in, this is passed to the ListItem, and the callsite chooses how to merge classnames?

dandy ✨

cheapsteak:
Thanks for going the extra mile to cater to support users being able to open new tabs :pray:

src/Select/index.tsx
justinanastos: Exactly right. The client will need to handle merging the class names. That function just needs to know to use `cx` instead of `classnames` if merging with something else from emotion.

src/Select/index.tsx
justinanastos: Values are _not_ guaranteed to be unique; and I think the resulting bug will produce the correct behavior if you give duplicates values.

apollo-bot2:
:rocket: PR was released in [`v8.12.0`](https://github.com/apollographql/space-kit/releases/tag/v8.12.0) :rocket:

PR #308, opened 2021-02-03
fix(Modal): Fix repeating animations with custom components

justinanastos:
Re-running `motion.custom(type)` on every render results in a new component on every render. When we render this with `React.createElement` (which is what `<MotionComponent>` does under the hood), this will cause the `<MotionComponent>` to destroy and create a new DOM element on every render. Why? `React.createElement` verifies that the component being rendered (`type`) is the same from render-to-render before doing a DOM diff to see if there is a change. In other words, if the `type` changes from one render to the next, React will assume you don't want to re-use the existing DOM and will destroy and re-create it.

I installed the canary in Studio UI to see if would fix the e2e and unit tests; it does! https://app.circleci.com/pipelines/github/mdg-private/studio-ui/26973/workflows/4323c9e1-9267-4fa6-8a00-a9055b5a6907

The solution? Simply memoize `MotionType` based on `type`.

<details>
  <summary>📦 Published PR as canary version: <code>8.13.2-canary.308.8001.0</code></summary>
  <br />
  
  ```bash
  npm install @apollo/space-kit@8.13.2-canary.308.8001.0
  # or 
  yarn add @apollo/space-kit@8.13.2-canary.308.8001.0
  ```
</details>

### DIFF
--- a/src/Modal/Modal.tsx
+++ b/src/Modal/Modal.tsx
@@ -180,8 +180,13 @@ export const Modal: React.FC<Props> = ({
    * We have to strip the type because we'll get an error about the union being
    * too complex to model because `motion` has over 100 options.
    */
-  const MotionComponent: React.ComponentType<MotionProps> =
-    typeof type === "string" ? motion[type] : motion.custom<MotionProps>(type);
+  const MotionComponent: React.ComponentType<MotionProps> = React.useMemo(
+    () =>
+      typeof type === "string"
+        ? motion[type]
+        : motion.custom<MotionProps>(type),
+    [type],
+  );
 
   return (
     <ClassNames>


cheapsteak (approved):
nice find 👍 
relatively safe to assume `type` is stable?

justinanastos:
> nice find 👍
> relatively safe to assume `type` is stable?

Yes @cheapsteak . When rendering something like `<Form />`, type is `Form`, which is stable. If we had something like `as={() => ...}`, then `type` would not be stable. Fortunately for us, `as` is a `React.ReactElement`, so we can rely on it to be stable. If we run into somewhere this bites us, we can provide a developer warning if `type` changes and require the consumer to memoize their element themselves.

https://github.com/apollographql/space-kit/blob/e12787314e5abafd2c016cac785e6c45a57b5a29/src/Modal/Modal.tsx#L27

Jephuff (approved):
🤦  yep! that makes sense! great find! :)

apollo-bot2:
:rocket: PR was released in [`v8.13.2`](https://github.com/apollographql/space-kit/releases/tag/v8.13.2) :rocket:

PR #310, opened 2021-02-09
[minor] Feature Intro Component

mayakoneval:
**Context**
In preparation for adding a Feature Intro component into Dev Graphs in Studio, I added it here first. This implementation mirrors the `FormControl` components that already exist in `space-kit`. That pattern is that there is a parent component called `---Control` that manages layout for all children components. In this case that is a heading, context, image and a couple buttons. Every component can be rendered on it's own, but it will be formatted well when wrapped with a `FeatureIntroControl`.

Figma: https://www.figma.com/file/BRioV3yezeo4PalvSf8Aju/Dev-Graphs?node-id=0%3A19029
Doc: https://apollographql.quip.com/tQa9AvdUA98j/Feature-Intro-component

**Notes/Questions**
This is a more complex implementation than is probably needed for this component right now. I chose to follow the `FormControl` pattern because it made sense for the modularity of the component, but it definitely could have been a simpler implementation. 

I chose to make the `FeatureIntroLearnMoreLink` have all link props rather than the `FeatureIntroControl` component. It feels more natural to me/this component is strict as a link.

Question: I can't spread all `props:HTMLAnchorProps` for the `<a` in `FeatureIntroLearnMoreLink` because the `useMemo` means there is infinite state updating. Taking any suggestions to be able to do this, since I think this would be ideal.

**Review**

Unfortunately I didn't break this up by commit. It is a direct match for the `FormControl` components, so might be helpful to look at that. 

**Testing**

I wrote some tests to make sure everything basically renders in `FeatureIntroControl.spec.tsx`. I didn't write tests for `FeatureIntroContext` because there is no external hook used by the children in `FeatureIntro`. 

<details>
  <summary>📦 Published PR as canary version: <code>8.15.1-canary.310.8285.0</code></summary>
  <br />
  
  ```bash
  npm install @apollo/space-kit@8.15.1-canary.310.8285.0
  # or 
  yarn add @apollo/space-kit@8.15.1-canary.310.8285.0
  ```
</details>


src/FeatureIntroLearnMoreLink/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Component to render Feature Intro learn more button.
+ *
+ * If this component is rendered in the children of `FeatureIntroControl`, then
+ * `FeatureIntroControl` will render this element in it's layout. Otherwise, it's
+ * rendered as-is.
+ */
+export const FeatureIntroLearnMoreLink: React.FC<React.HTMLProps<
+  HTMLAnchorElement
+>> = ({ className, style, id, href, rel, target }) => {
```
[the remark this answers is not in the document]
Jephuff: hmm yeah, we would have to do some work around to make it stable.
Perhaps this could accept an `as` prop where someone could do `as={<a ...any prop they want />}` that would also allow us to make it a react-router link if we need that in the future

src/shared/FeatureIntroControlContext/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+ */
+interface FeatureIntroControlContextConfig {
+  id?: string;
+}
+
+/**
+ * Required: dismiss, heading, content
+ */
+interface FeatureIntroControl extends FeatureIntroControlContextConfig {
+  heading: React.ReactNode | undefined;
```
Jephuff, line 14: NBD `React.ReactNode` includes `undefined` so these can just be `React.ReactNode`
```suggestion
  heading: React.ReactNode;
```

src/shared/FeatureIntroControlContext/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+    | React.Dispatch<React.SetStateAction<FeatureIntroControl["heading"]>>
+    | undefined;
+  content: React.ReactNode | undefined;
+  setContent:
+    | React.Dispatch<React.SetStateAction<FeatureIntroControl["content"]>>
+    | undefined;
+  image: React.ReactNode | undefined;
+  setImage:
+    | React.Dispatch<React.SetStateAction<FeatureIntroControl["image"]>>
+    | undefined;
```
Jephuff, line 25: Oh it looks like the context is either the interface or undefined, so I think we can require the setters here
```suggestion
  setImage: React.Dispatch<React.SetStateAction<FeatureIntroControl["image"]>>;
```

src/shared/FeatureIntroControlContext/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+    dismissButton,
+    setDismissButton,
+    headingId,
+    contentId,
+    imageId,
+    learnMoreLinkId,
+    dismissButtonId,
+  } = React.useContext(FeatureIntroControlContext) || {};
+
+  return React.useMemo(
```
Jephuff, line 107: Can we move the useMemo into the context provider? that will make things render a little less but also means we only need to define these keys once 👍

src/shared/FeatureIntroControlContext/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  return (
+    <FeatureIntroControlContext.Provider
+      value={{
+        id,
+        contentId: id && `${id}-content`,
+        headingId: id && `${id}-heading`,
+        imageId: id && `${id}-image`,
+        learnMoreLinkId: id && `${id}-learn-more-link`,
+        dismissButtonId: id && `${id}-dismiss-button`,
```
Jephuff, line 66: NBD I might not have all these in the provider, each child can generate there own Id based on the `id` in the context which would reduce a bit of the noise here :)
```suggestion
```

src/FeatureIntroControl/InteractiveFeatureIntro.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { colors } from "../colors";
+import Tippy from "@tippyjs/react";
+import { FeatureIntroControl } from "../FeatureIntroControl";
+import { FeatureIntroContent } from "../FeatureIntroContent";
+import { FeatureIntroHeading } from "../FeatureIntroHeading";
+import { FeatureIntroLearnMoreLink } from "../FeatureIntroLearnMoreLink";
+import { FeatureIntroDismissButton } from "../FeatureIntroDismissButton";
+import { FeatureIntroTippyStyles } from "./FeatureIntroTippyStyles";
+
+export const InteractiveFeatureIntro: React.FC = () => {
```
Jephuff, line 13: This component looks like it maybe belongs in studio code to me?

src/FeatureIntroControl/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+                paddingBottom: 20,
+                flexBasis: image ? "50%" : "auto",
+              }}
+            >
+              <div css={{ paddingBottom: 12 }}>
+                {React.isValidElement<any>(heading) &&
+                  React.cloneElement(heading, {
+                    ...heading.props,
+                    id: headingId,
+                  })}
```
Jephuff, line 164: The heading component will already have the id set by its self I think so this this clone isn't needed maybe?
```suggestion
                {heading}
```

Probably same thing applies to the other spots where the id is being added?

Jephuff (commented):
🎉

src/shared/FeatureIntroControlContext/index.tsx
mayakoneval: If these elements are called outside of the Context they will be undefined and we want to return the raw element, but every child calls `useFeatureIntroControlInternalContext` to see if they are defined before returning themselves, I think thats why I had these as undefined.

src/FeatureIntroControl/InteractiveFeatureIntro.tsx
mayakoneval: This was more of a test for me to make sure things work since I had a lot of issues npm linking. I'm happy to delete it, but I also think it shows how the FeatureIntro is supposed to work (in a Popover). Let me know!

src/shared/FeatureIntroControlContext/index.tsx
Jephuff: ahh yep that totally makes sense, how would you feel about `useFeatureIntroControlInternalContext` returning undefined when not in a context provider, which would change usage a bit to something like
```
const featureContext = useFeatureIntroControlInternalContext()

const id = featureContext?.id
... make the element
return featureContext ? null : element
```

I don't have strong feelings on that, so I'll definitely leave it up to you :)

src/shared/FeatureIntroControlContext/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+          setLearnMoreLink,
+          dismissButton,
+          setDismissButton,
+        }}
+      >
+        {children}
+      </FeatureIntroControlContext.Provider>
+    ),
+    [children, content, dismissButton, heading, id, image, learnMoreLink],
+  );
```
Jephuff, line 74: This piece can just memoize the `data` for the context 👍 
```suggestion
  const contextValue: FeatureIntroControl = React.useMemo(
    () => (
      {
          id,
          content,
          setContent,
          heading,
          setHeading,
          image,
          setImage,
          learnMoreLink,
          setLearnMoreLink,
          dismissButton,
          setDismissButton,
        }
    ),
    [content, dismissButton, heading, id, image, learnMoreLink],
  );
  return <FeatureIntroControlContext.Provider value={contextValue}>
    {children}
  </FeatureIntroControlContext.Provider>      
```

src/shared/FeatureIntroControlContext/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+    setHeading,
+    content,
+    setContent,
+    image,
+    setImage,
+    learnMoreLink,
+    setLearnMoreLink,
+    dismissButton,
+    setDismissButton,
+  };
```
Jephuff, line 107: With the useMemo in the provider, this can just return the context
```suggestion
  return React.useContext(FeatureIntroControlContext) || {};
```
though now I am realizing that we need to memoize that default `{}` so if you dont go with that other option of return undefined from the hook this would need to be
```suggestion
  const contextValue = React.useContext(FeatureIntroControlContext);
  return useMemo(() => contextValue ?? {}, [contextValue])
```

src/FeatureIntroLearnMoreLink/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * rendered as-is.
+ */
+export const FeatureIntroLearnMoreLink: React.FC<
+  React.HTMLProps<HTMLAnchorElement> | Props
+> = (id) => {
+  const { setLearnMoreLink } = useFeatureIntroControlInternalContext();
+
+  const element = React.useMemo(
+    () => (
+      <div id={id && `${id}-learn-more-link`}>
```
Jephuff, line 27: ahh I think these will need to still use the root id for the generated value
```suggestion
  const { setLearnMoreLink, id: contextId } = useFeatureIntroControlInternalContext();

  const element = React.useMemo(
    () => (
      <div id={id ?? contextId && `${contextId}-learn-more-link`}>
```

src/FeatureIntroControl/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+              borderColor: colors.white,
+              borderRadius: 4,
+              display: "flex",
+              flexDirection,
+            }}
+          >
+            {React.isValidElement(image) &&
+              React.cloneElement(image, {
+                ...image.props,
+              })}
```
Jephuff, line 149: I believe the cloning here isn't necessary 
```suggestion
            {image}
```

src/FeatureIntroLearnMoreLink/index.tsx
mayakoneval: 🤦

mayakoneval:
Is it normal for some storybook tests to change slightly and need to be re-accepted? Or did I screw something up

justinanastos:
> Is it normal for some storybook tests to change slightly and need to be re-accepted? Or did I screw something up

@mayakoneval There are two tests that keep changing, you'll know them because they have debugging info right in the story. These changes are because you're' not up to date on `main`. Rebase and you'll see the Modal stories match back up

Jephuff (approved):

apollo-bot2:
:rocket: PR was released in [`v8.16.0`](https://github.com/apollographql/space-kit/releases/tag/v8.16.0) :rocket:

PR #317, opened 2021-02-20
Update to emotion 11

hwillson:
This PR updates to emotion 11, following their [recommended migration steps](https://emotion.sh/docs/emotion-11). I've had to bump the major version of `@apollo/space-kit`, since we're recommending people install emotion as a peer dep, which means requiring the new `@emotion/react` package is a breaking change. Thanks!


jest.config.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
   },
   preset: "ts-jest/presets/js-with-babel",
   setupFilesAfterEnv: ["<rootDir>jest/setup.js"],
-  snapshotSerializers: ["jest-emotion"],
+  snapshotSerializers: ["@emotion/jest/serializer"],
```
hwillson, line 18: Note that the `jest-emotion` package can no longer be used with emotion 11.

src/Button/Button.tsx
```diff
@@ -3,7 +3,7 @@
 import { colors, ShadedColor } from "../colors";
 import type { Property, SimplePseudos } from "csstype";
 import { base } from "../typography";
-import { ClassNames, jsx, ObjectInterpolation } from "@emotion/core";
+import { ClassNames, jsx, CSSObject } from "@emotion/react";
```
hwillson, line 6: `ObjectInterpolation` is no longer re-exported from the `@emotion/react` package. It is available through `@emotion/serialize` (which is a dep of `@emotion/react`), but I didn't want to depend on a child dep, or ask people to include `@emotion/serialize` as a peer dep. That being said, the only difference between `ObjectInterpolation` and `CSSObject` is that `ObjectInterpolation` accepts generics, but in this case we were using a generic of `undefined` anyways, so it shouldn't matter.

justinanastos: 👍

justinanastos (commented):
Sorry it took me so long to get to this; it looks great @hwillson . There was one regression related to how the `css` prop worked that I've resolved.

I'm going to merge this after https://github.com/mdg-private/studio-ui/pull/3847 is merged so we don't block any Space Kit updates for Studio UI.

PR #326, opened 2021-03-05
Expose color prop for elect trigger

jgzuke:
Personally I find in unintuitive that the `color`, `feel`, `size`, etc passed on the `triggerAs` get overwritten. I went with this as the smallest unblocking change, but also happy to switch this to follow props passed to `triggerAs` so we dont need a mix of top level props and `triggerAs` to style a custom trigger.

<details>
  <summary>📦 Published PR as canary version: <code>8.18.1-canary.326.8945.0</code></summary>
  <br />
  
  ```bash
  npm install @apollo/space-kit@8.18.1-canary.326.8945.0
  # or 
  yarn add @apollo/space-kit@8.18.1-canary.326.8945.0
  ```
</details>

### DIFF
--- a/src/Select/index.tsx
+++ b/src/Select/index.tsx
@@ -143,7 +143,7 @@ interface Props
     >,
     Pick<
       React.ComponentProps<typeof Button>,
-      "aria-labelledby" | "aria-describedby" | "feel" | "style"
+      "aria-labelledby" | "aria-describedby" | "feel" | "style" | "color"
     >,
     Pick<
       React.ComponentProps<typeof ListConfigProvider>,
@@ -265,6 +265,7 @@ export const Select: React.FC<Props> = ({
   defaultValue,
   disabled = false,
   feel,
+  color,
   labelPropsCallbackRef,
   listAs = <List startIconAs={<div css={{ alignSelf: "baseline" }} />} />,
   margin = "auto",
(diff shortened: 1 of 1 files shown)


src/Select/index.tsx
```diff
@@ -265,6 +265,7 @@ export const Select: React.FC<Props> = ({
   defaultValue,
   disabled = false,
   feel,
+  color,
```
justinanastos, line 268: I feel like it'd be more intuitive to set the default here
```suggestion
  color = colors.white,
```

src/Select/index.tsx
```diff
@@ -533,7 +534,7 @@ export const Select: React.FC<Props> = ({
                   React.isValidElement(triggerAs) &&
                     (triggerAs.props as any).className,
                 ),
-                color: colors.white,
+                color: color ?? colors.white,
```
justinanastos, line 537: And then just read `color`
```suggestion
                color,
```

justinanastos (approved):
Made a suggestion but it's really about ergonomics and doesn't affect the behavior; your call if you want to take the suggestion or merge as is. Thanks for taking the time to add this! 👍

apollo-bot2:
:rocket: PR was released in [`v8.19.0`](https://github.com/apollographql/space-kit/releases/tag/v8.19.0) :rocket:

PR #330, opened 2021-03-10
AR-2148 Fix repeated rendering of global styles for tooltip-like components

justinanastos:
Resolves issue AR-2148

We are rendering `<Global>` styles for all instances of `AbstractTooltip`, of which there can be thousands lying in wait on a single page. I incorrectly believed that these would be rendered just once to the DOM, but I was wrong; they are rendered once per instance rendered. This resulted in thousands of duplicated `<style>` tags being added.

## Release Notes

* Create an internal component called `SingletonComponent` (the name is poor and can be changed whenever we want because it's internal) that will take a string `identity` and only allow a single instance in the DOM regardless of how many times it's rendered. This requires us to use `SpaceKitProvider`. Tooltips can be used without `SpaceKitProvider` and will warn the user in the console.
* Wrap the two `<Global>` styles with the `SingletonComponent` to remove all excess DOM `<style />`s

<details>
  <summary>📦 Published PR as canary version: <code>8.17.4-canary.330.8834.0</code></summary>
  <br />
  
  ```bash
  npm install @apollo/space-kit@8.17.4-canary.330.8834.0
  # or 
  yarn add @apollo/space-kit@8.17.4-canary.330.8834.0
  ```
</details>

### DIFF
--- a/src/AbstractTooltip/abstractTooltip/TippyStyles.tsx
+++ b/src/AbstractTooltip/abstractTooltip/TippyStyles.tsx
@@ -3,38 +3,46 @@ import React from "react";
 import { base } from "../../typography";
 import { colors } from "../../colors";
 import { Global, css } from "@emotion/core";
+import { SingletonComponent } from "../../shared/components/SingletonComponent";
 
 export const TippyStyles: React.FC = () => (
-  <Global
-    styles={css({
-      ".tippy-box": {
-        "&[data-theme=space-kit]": {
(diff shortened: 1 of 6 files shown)


cheapsteak (approved):
> I incorrectly believed that these would be rendered just once to the DOM

ack, apologies, ditto, I figured since emotion was using content-hash to make classnames it would be able to avoid creating duplicate style tags  

🚢

PR #354, opened 2021-06-07
Add warning circle icon

esilverm:
This PR adds a new StreamLine icon to indicate "Checks Failed with Broken Operations" in a launch sequence component.

The icon's name may need to be adjusted. Let me know if that is the case

Screen Shot 2021-06-07 at 1 20 56 PM


mayakoneval (approved):
🚀

jglovier:
Is there any reason we can't use the existing warning or error indicator icons for this use case? I'm concerned that introducing this icon may dilute the signal of those, and be a bit confusing why we have so many different styles for warning/error/etc.

PR #373, opened 2021-08-23
Modal: no title, no description dont render extra margin-top

cy:
For https://github.com/mdg-private/studio-ui/pull/4669, a modal with no title and no description does not need the `margin-top` added to space the content from the title and description.

## Before
Screen Shot 2021-08-23 at 4 50 29 PM

## After
Screen Shot 2021-08-23 at 4 50 41 PM

<details>
  <summary>📦 Published PR as canary version: <code>9.6.1-canary.373.9851.0</code></summary>
  <br />
  
  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @apollo/space-kit@9.6.1-canary.373.9851.0
  # or 
  yarn add @apollo/space-kit@9.6.1-canary.373.9851.0
  ```
</details>

PR #374, opened 2021-08-24
modal - should be no title OR description on top margin

cy:
I made a mistake in https://github.com/apollographql/space-kit/pull/373, instead of not having a top margin between the content if there's no title and no description, it should be no title OR no description. The mistake's effect can see in this chromatic https://www.chromatic.com/test?appId=5a77a6d38a9e3c002045d20d&id=61255022235bb7003aae37f0

before
image

after
Screen Shot 2021-08-24 at 5 16 41 PM

### DIFF
--- a/src/Modal/Modal.tsx
+++ b/src/Modal/Modal.tsx
@@ -285,11 +285,10 @@ export const Modal: React.FC<Props> = ({
             </div>
             <div
               css={css(
-                title &&
-                  description && {
-                    marginTop:
-                      size === "large" ? 24 : size === "medium" ? 16 : 12,
-                  },
+                (title || description) && {
+                  marginTop:
+                    size === "large" ? 24 : size === "medium" ? 16 : 12,
+                },
                 verticalScrollMode === "children" && {
                   overflowY: "auto",
                 },


src/Modal/Modal.tsx
```diff
@@ -285,11 +285,10 @@ export const Modal: React.FC<Props> = ({
             </div>
             <div
               css={css(
-                title &&
-                  description && {
-                    marginTop:
-                      size === "large" ? 24 : size === "medium" ? 16 : 12,
-                  },
+                (title || description) && {
```
jgzuke, line 288: Im not quite sure about having description here? Does the description render in the top section?

jgzuke (approved):
One question that might be worth double checking 😄

src/Modal/Modal.tsx
cheapsteak: I think the logic is right but the title is possibly confused?

this makes sense `(title || description) ? defaultMargin : noMargin`

(gosh this component is getting tangly though, really want to chop it up)

cheapsteak: > Does the description render in the top section


it looks like it does
image

cheapsteak (approved):
🚢

src/Modal/Modal.tsx
cy: yeah, the description does render in the top section - very tangly component FOR SURE

apollo-bot2:
:rocket: PR was released in [`v9.6.2`](https://github.com/apollographql/space-kit/releases/tag/v9.6.2) :rocket:

PR #375, opened 2021-08-30
allow ConfirmationTooltip to have an initial tip

cy:
This is to allow for @jglovier's request in https://github.com/mdg-private/studio-ui/pull/4813#discussion_r697719047 to allow for the confirmation tooltip to show an initial message "Copy to clipboard" on hover before the "Copied" tip on click.

Currently `Tooltip` and `ConfirmationTooltip` overwrite each other, the inner most nested tip will take effect, and the outer one will have no effect. So I've attempted to allow for this behaviour by adding an `initialContent` prop to `ConfirmationTooltip`. Onclick, the displayed content switches to the default content, on hide it switches back to initial content. This works well, except for a split second when the tip is hiding the initial text can be seen again. Setting another timeout before doing `setDisplayedContent(initalContent);` in `onHide` doesn't seem to work. Anyone have an idea on how to solve this? 😖

CPT2108301236-254x143

### DIFF
--- a/src/ConfirmationTooltip/index.tsx
+++ b/src/ConfirmationTooltip/index.tsx
@@ -6,7 +6,9 @@ type Props = Pick<
   React.ComponentProps<typeof AbstractTooltip>,
   "children" | "content" | "disabled"
 > &
-  AbstractTooltipProps;
+  AbstractTooltipProps & {
+    initalContent?: string;
+  };
 
 /**
  * Display a non-interactive tooltip after an element is clicked on and then
@@ -14,6 +16,8 @@ type Props = Pick<
  */
 export const ConfirmationTooltip: React.FC<Props> = ({
   children,
+  initalContent,
+  content,
   ...props
 }) => {
   /**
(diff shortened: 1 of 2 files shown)


cheapsteak:
there are two places in studio-ui that use a `TooltipWithMutableContent`, where content is reset on mouseenter to avoid the flicker

src/ConfirmationTooltip/index.tsx
```diff
@@ -32,28 +36,39 @@ export const ConfirmationTooltip: React.FC<Props> = ({
...
-        if (timeoutRef.current) {
-          window.clearTimeout(timeoutRef.current);
-        }
+  const [displayedContent, setDisplayedContent] = React.useState(
+    initalContent ?? content,
+  );
 
-        timeoutRef.current = null;
+  return (
+    <div
```
cheapsteak, line 44: might adding a div wrapper cause unexpected layout problems?  

(my understanding is that this component did not previously generate DOM elements)?

Ideally if possible we could make our components headless (in the vein of downshift and headless-ui)

cy:
> there are two places in studio-ui that use a `TooltipWithMutableContent`, where content is reset on mouseenter to avoid the flicker

This is exactly what I needed!! Thank you 🙇‍♀️

PR #379, opened 2021-09-22
Update tooltips to support controlled visibility

Jephuff:
Right now we do have `forceVisibleForTestingOnly` which does work if you want to make a tooltip always visible, but it creates a bunch of errors, doesn't work to keep it closed and of course the name implies you shouldn't use it. This adds official support for the `visible` prop so we can have fully controlled tooltips


cheapsteak (approved):

apollo-bot2:
:rocket: PR was released in [`v9.9.0`](https://github.com/apollographql/space-kit/releases/tag/v9.9.0) :rocket:

PR #385, opened 2022-01-19
Add deprecated message to readme

Jephuff:
The space-kit code is being moved into an internal repository, this adds a deprecated message for anyone landing here.


cheapsteak (approved):
There's also an "archived" mode toggle somewhere, I think?

Jephuff:
> There's also an "archived" mode toggle somewhere, I think?

I don't have access to the settings, so I'll merge this for now but definitely sounds like something that would be good to turn on :D

